PR #5, opened 2023-03-31
Add ISCE2 and burst downloading workflow

forrestfwilliams:
This PR adds the ability to download burst data/metadata from ASF's burst extractor, and package it as a psuedo-SAFE archive. It also adds the ability to call ISCE2's `topsApp.py`.

### DIFF
--- a/src/hyp3_isce2/__main__.py
+++ b/src/hyp3_isce2/__main__.py
@@ -6,7 +6,6 @@ import logging
 from argparse import ArgumentParser
 
 from hyp3_isce2.process import topsapp_burst
-
 from hyp3lib.aws import upload_file_to_s3
 from hyp3lib.image import create_thumbnail
 
--- a/src/hyp3_isce2/__main__.py
+++ b/src/hyp3_isce2/__main__.py
@@ -1,14 +1,14 @@
 """
 ISCE2 processing for HyP3
 """
+import copy
 import logging
 from argparse import ArgumentParser
 
 from hyp3lib.aws import upload_file_to_s3
 from hyp3lib.image import create_thumbnail
 
-
-from hyp3_isce2.process import process_isce2
+from hyp3_isce2.process import topsapp_burst
 
 
 def main():
@@ -18,19 +18,25 @@ def main():
     parser = ArgumentParser()
     parser.add_argument('--bucket', help='AWS S3 bucket HyP3 for upload the final product(s)')
     parser.add_argument('--bucket-prefix', default='', help='Add a bucket prefix to product(s)')
-
-    # TODO: Your arguments here
-    parser.add_argument('--greeting', default='Hello world!',
-                        help='Write this greeting to a product file')
+    parser.add_argument('--reference-scene', type=str, required=True)
+    parser.add_argument('--secondary-scene', type=str, required=True)
+    parser.add_argument('--swath-number', type=int, required=True)
+    parser.add_argument('--polarization', type=str, default='VV')
+    parser.add_argument('--reference-burst-number', type=int, required=True)
+    parser.add_argument('--secondary-burst-number', type=int, required=True)
+    parser.add_argument('--azimuth-looks', type=int, default=4)
+    parser.add_argument('--range-looks', type=int, default=20)
 
     args = parser.parse_args()
 
-    logging.basicConfig(format='%(asctime)s - %(levelname)s - %(message)s',
-                        datefmt='%m/%d/%Y %I:%M:%S %p', level=logging.INFO)
-
-    product_file = process_isce2(
-        greeting=args.greeting,
+    logging.basicConfig(
+        format='%(asctime)s - %(levelname)s - %(message)s', datefmt='%m/%d/%Y %I:%M:%S %p', level=logging.INFO
     )
+
+    args_dict = copy.copy(args.__dict__)
+    del args_dict['bucket']
+    del args_dict['bucket_prefix']
+    product_file = topsapp_burst(**args_dict)
 
     if args.bucket:
         upload_file_to_s3(product_file, args.bucket, args.bucket_prefix)
--- a/src/hyp3_isce2/__main__.py
+++ b/src/hyp3_isce2/__main__.py
@@ -1,7 +1,6 @@
 """
 ISCE2 processing for HyP3
 """
-import copy
 import logging
 from argparse import ArgumentParser
 
--- a/src/hyp3_isce2/__main__.py
+++ b/src/hyp3_isce2/__main__.py
@@ -5,9 +5,10 @@ import copy
 import logging
 from argparse import ArgumentParser
 
-from hyp3_isce2.process import topsapp_burst
 from hyp3lib.aws import upload_file_to_s3
 from hyp3lib.image import create_thumbnail
+
+from hyp3_isce2.process import topsapp_burst
 
 
 def main():
--- a/src/hyp3_isce2/__main__.py
+++ b/src/hyp3_isce2/__main__.py
@@ -33,10 +33,16 @@ def main():
         format='%(asctime)s - %(levelname)s - %(message)s', datefmt='%m/%d/%Y %I:%M:%S %p', level=logging.INFO
     )
 
-    args_dict = copy.copy(args.__dict__)
-    del args_dict['bucket']
-    del args_dict['bucket_prefix']
-    product_file = topsapp_burst(**args_dict)
+    product_file = topsapp_burst(
+        reference_scene=args.reference_scene,
+        secondary_scene=args.secondary_scene,
+        swath_number=args.swath_number,
+        polarization=args.polarization,
+        reference_burst_number=args.reference_burst_number,
+        secondary_burst_number=args.secondary_burst_number,
+        azimuth_looks=args.azimuth_looks,
+        range_looks=args.range_looks,
+    )
 
     if args.bucket:
         upload_file_to_s3(product_file, args.bucket, args.bucket_prefix)
--- a/src/hyp3_isce2/__main__.py
+++ b/src/hyp3_isce2/__main__.py
@@ -5,10 +5,10 @@ import copy
 import logging
 from argparse import ArgumentParser
 
+from hyp3_isce2.process import topsapp_burst
+
 from hyp3lib.aws import upload_file_to_s3
 from hyp3lib.image import create_thumbnail
-
-from hyp3_isce2.process import topsapp_burst
 
 
 def main():
--- a/src/hyp3_isce2/burst.py
+++ b/src/hyp3_isce2/burst.py
@@ -233,5 +233,6 @@ def download_bursts(param_list: Iterator[BurstParams]) -> List[BurstMetadata]:
 
 
 if __name__ == '__main__':
-    burst_params1 = BurstParams('S1A_IW_SLC__1SDV_20230321T001351_20230321T001418_047735_05BBF8_2C3A', 'IW2', 'VV', 6)
-    metadata = download_bursts([burst_params1])
+    burst_params1 = BurstParams('S1A_IW_SLC__1SDV_20211229T231926_20211229T231953_041230_04E66A_3DBE', 'IW2', 'VV', 2)
+    burst_params2 = BurstParams('S1A_IW_SLC__1SDV_20220110T231926_20220110T231953_041405_04EC57_103E', 'IW2', 'VV', 2)
+    metadata = download_bursts([burst_params1, burst_params2])
--- a/src/hyp3_isce2/burst.py
+++ b/src/hyp3_isce2/burst.py
@@ -160,7 +160,7 @@ def spoof_safe(asf_session: requests.Session, burst: BurstMetadata, base_path: P
             ├── calibration.xml
             └── noise.xml
     """
-    safe_path = base_path / burst.safe_name
+    safe_path = base_path / f'{burst.safe_name}.SAFE'
     annotation_path = safe_path / 'annotation'
     calibration_path = safe_path / 'annotation' / 'calibration'
     measurement_path = safe_path / 'measurement'
@@ -235,6 +235,9 @@ def download_bursts(param_list: Iterator[BurstParams]) -> List[BurstMetadata]:
 
 
 if __name__ == '__main__':
-    burst_params1 = BurstParams('S1A_IW_SLC__1SDV_20211229T231926_20211229T231953_041230_04E66A_3DBE', 'IW2', 'VV', 2)
-    burst_params2 = BurstParams('S1A_IW_SLC__1SDV_20220110T231926_20220110T231953_041405_04EC57_103E', 'IW2', 'VV', 2)
+    burst_params1 = BurstParams('S1A_IW_SLC__1SDV_20211229T231926_20211229T231953_041230_04E66A_3DBE', 'IW2', 'VV', 3)
+    burst_params2 = BurstParams('S1A_IW_SLC__1SDV_20220110T231926_20220110T231953_041405_04EC57_103E', 'IW2', 'VV', 3)
     metadata = download_bursts([burst_params1, burst_params2])
+    # with get_asf_session() as asf_session:
+    #     download_metadata(asf_session, burst_params1, 'reference.xml')
+    #     download_metadata(asf_session, burst_params2, 'secondary.xml')
--- a/src/hyp3_isce2/burst.py
+++ b/src/hyp3_isce2/burst.py
@@ -7,7 +7,9 @@ from pathlib import Path
 from typing import Iterator, List, Tuple, Union
 
 import pandas as pd
+
 import requests
+
 from shapely import geometry
 
 URL = 'https://sentinel1-burst.asf.alaska.edu'
--- a/src/hyp3_isce2/burst.py
+++ b/src/hyp3_isce2/burst.py
@@ -236,6 +236,3 @@ if __name__ == '__main__':
     burst_params1 = BurstParams('S1A_IW_SLC__1SDV_20211229T231926_20211229T231953_041230_04E66A_3DBE', 'IW2', 'VV', 3)
     burst_params2 = BurstParams('S1A_IW_SLC__1SDV_20220110T231926_20220110T231953_041405_04EC57_103E', 'IW2', 'VV', 3)
     metadata = download_bursts([burst_params1, burst_params2])
-    # with get_asf_session() as asf_session:
-    #     download_metadata(asf_session, burst_params1, 'reference.xml')
-    #     download_metadata(asf_session, burst_params2, 'secondary.xml')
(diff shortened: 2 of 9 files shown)


src/hyp3_isce2/__main__.py
```diff
@@ -1,37 +1,43 @@
...
-    product_file = process_isce2(
-        greeting=args.greeting,
+    logging.basicConfig(
+        format='%(asctime)s - %(levelname)s - %(message)s', datefmt='%m/%d/%Y %I:%M:%S %p', level=logging.INFO
     )
 
+    args_dict = copy.copy(args.__dict__)
+    del args_dict['bucket']
+    del args_dict['bucket_prefix']
+    product_file = topsapp_burst(**args_dict)
```
jtherrmann, line 39: Gross, but effective :)

src/hyp3_isce2/burst.py
```diff
@@ -0,0 +1,243 @@
...
+    return bursts
+
+
+if __name__ == '__main__':
+    burst_params1 = BurstParams('S1A_IW_SLC__1SDV_20211229T231926_20211229T231953_041230_04E66A_3DBE', 'IW2', 'VV', 3)
+    burst_params2 = BurstParams('S1A_IW_SLC__1SDV_20220110T231926_20220110T231953_041405_04EC57_103E', 'IW2', 'VV', 3)
+    metadata = download_bursts([burst_params1, burst_params2])
+    # with get_asf_session() as asf_session:
+    #     download_metadata(asf_session, burst_params1, 'reference.xml')
+    #     download_metadata(asf_session, burst_params2, 'secondary.xml')
```
jtherrmann, line 243: Are these comments still needed?

src/hyp3_isce2/__main__.py
forrestfwilliams: yeah, definetly agree. Do you think it would be better to write them all out? Open to suggestions!

src/hyp3_isce2/burst.py
forrestfwilliams: nope, I'll remove them.

forrestfwilliams:
Trufflehog is failing on some checksums included in the example xml files we need for testing. Not sure if there's anything we can do about that.

src/hyp3_isce2/__main__.py
jtherrmann: Whenever I need to pass most/all of the cli args to a function, I just write them all out as keyword arguments. I do think it is slightly more readable that way.

src/hyp3_isce2/burst.py
```diff
@@ -0,0 +1,238 @@
...
+        centroid = tuple([x[0] for x in footprint.centroid.xy])
+        return footprint, footprint.bounds, centroid
+
+
+def create_burst_request(params: BurstParams, content: str) -> dict:
+    """
+    Syntax: www.API-URL.com/<granule>/<subswath>/<pol>/<burst_number>.(xml|tiff)
+    """
+    filetypes = {'metadata': 'xml', 'geotiff': 'tiff'}
+    exstension = filetypes[content]
```
jtherrmann, line 95:
```suggestion
    extension = filetypes[content]
```

src/hyp3_isce2/burst.py
```diff
@@ -0,0 +1,238 @@
...
+    burst_number_zero_indexed = params.burst_number - 1
+    url = f'{URL}/{params.granule}/{params.swath}/{params.polarization}/{burst_number_zero_indexed}.{exstension}'
```
jtherrmann, line 97:
```suggestion
    url = f'{URL}/{params.granule}/{params.swath}/{params.polarization}/{burst_number_zero_indexed}.{extension}'
```

src/hyp3_isce2/burst.py
```diff
@@ -0,0 +1,238 @@
...
+        y = y1 + y2
+        footprint = geometry.Polygon(zip(x, y))
+        centroid = tuple([x[0] for x in footprint.centroid.xy])
+        return footprint, footprint.bounds, centroid
+
+
+def create_burst_request(params: BurstParams, content: str) -> dict:
+    """
+    Syntax: www.API-URL.com/<granule>/<subswath>/<pol>/<burst_number>.(xml|tiff)
+    """
```
jtherrmann, line 93: This is clear from the function body.
```suggestion
```

src/hyp3_isce2/burst.py
```diff
@@ -0,0 +1,238 @@
...
+        time.sleep(15)
+        return False
+
+    response.raise_for_status()
+    return True
+
+
+def download_from_extractor(asf_session: requests.Session, burst_params: BurstParams, content: str):
+    burst_request = create_burst_request(burst_params, content=content)
+    burst_request['cookies'] = {'asf-urs': asf_session.cookies['asf-urs']}
```
jtherrmann, line 112: I would suggest renaming `create_burst_request` to something like `get_burst_request_url` and have the function return `str` rather than `dict`, then do something like:

```suggestion
    burst_request = {
        'url': create_burst_request(burst_params, content=content),
        'cookies': {'asf-urs': asf_session.cookies['asf-urs']}
    }
```

src/hyp3_isce2/burst.py
```diff
@@ -0,0 +1,238 @@
...
+            break
+
+    if not downloaded:
+        raise RuntimeError('Download failed too many times')
+
+    return response.content
+
+
+def download_metadata(asf_session: requests.Session, burst_params: BurstParams, out_file: Union[Path, str] = None):
+    content = download_from_extractor(asf_session, burst_params, 'metadata')
```
jtherrmann, line 128: It confused me that we're setting a variable `content` equal to the return value of `download_from_extractor`, but `content` is also the name of a function parameter to `download_from_extractor` and `create_burst_request`. I suggest we rename those function parameters to something like `filetype` or `content_type` to distinguish them from the variable `content` declared here.

src/hyp3_isce2/burst.py
```diff
@@ -0,0 +1,238 @@
...
+        if downloaded:
+            break
+
+    if not downloaded:
+        raise RuntimeError('Download failed too many times')
+
+    return response.content
+
+
+def download_metadata(asf_session: requests.Session, burst_params: BurstParams, out_file: Union[Path, str] = None):
```
jtherrmann, line 127: What's the return type? Looks like it might be `Union[ET, str]`?

src/hyp3_isce2/burst.py
jtherrmann, line 1: We'll want return type annotations for all functions and methods.

src/hyp3_isce2/burst.py
```diff
@@ -0,0 +1,238 @@
...
+
+def spoof_safe(asf_session: requests.Session, burst: BurstMetadata, base_path: Path = Path('.')) -> Path:
+    """Creates this file structure:
+    SLC.SAFE/
+    ├── manifest.safe
+    ├── measurement/
+    │   └── burst.tif
+    └── annotation/
+        ├── annotation.xml
+        └── calbiration/
```
jtherrmann, line 157:
```suggestion
        └── calibration/
```

src/hyp3_isce2/burst.py
```diff
@@ -0,0 +1,238 @@
...
+            └── noise.xml
+    """
+    safe_path = base_path / f'{burst.safe_name}.SAFE'
+    annotation_path = safe_path / 'annotation'
+    calibration_path = safe_path / 'annotation' / 'calibration'
+    measurement_path = safe_path / 'measurement'
+    paths = [annotation_path, calibration_path, measurement_path]
+    for p in paths:
+        if not p.exists():
+            p.mkdir(parents=True)
```
jtherrmann, line 168:
```suggestion
        p.mkdir(parents=True, exist_ok=True)
```

src/hyp3_isce2/burst.py
```diff
@@ -0,0 +1,238 @@
...
+) -> Tuple[float, float, float, float]:
+    bbox1 = geometry.box(*poly1.bounds)
+    bbox2 = geometry.box(*poly2.bounds)
+    intersection = bbox1.intersection(bbox2)
+    bounds = intersection.bounds
+
+    x, y = (0, 1) if is_ascending else (2, 1)
+    roi = geometry.Point(bounds[x], bounds[y]).buffer(0.005)
+    bounds = roi.bounds  # returns (minx, miny, maxx, maxy)
+    return bounds
```
jtherrmann, line 195:
```suggestion
    return roi.bounds  # returns (minx, miny, maxx, maxy)
```

src/hyp3_isce2/burst.py
```diff
@@ -0,0 +1,238 @@
...
+    ET.ElementTree(burst.noise).write(calibration_path / burst.noise_name, **et_args)
+    ET.ElementTree(burst.manifest).write(safe_path / 'manifest.safe', **et_args)
+
+    burst_params = BurstParams(burst.safe_name, burst.swath, burst.polarization, burst.burst_number)
+    download_burst(asf_session, burst_params, measurement_path / burst.measurement_name)
+
+    return safe_path
+
+
+# TODO currently only validated for descending orbits
```
jtherrmann, line 183: Is this blocking?

src/hyp3_isce2/burst.py
```diff
@@ -0,0 +1,238 @@
...
+    For each burst:
+        1. Download metadata
+        2. Create BurstMetadata object
+        3. Create directory structure
+        4. Write metadata
+        5. Download and write geotiff
+    """
+    bursts = []
+    with get_asf_session() as asf_session:
+        for i, params in enumerate(param_list):
```
jtherrmann, line 223: Earlier a variable `ii` was used, would be nice to choose one or the other for the sake of consistency.

src/hyp3_isce2/burst.py
```diff
@@ -0,0 +1,238 @@
...
+
+    print('SAFEs created!')
+
+    return bursts
+
+
+if __name__ == '__main__':
+    burst_params1 = BurstParams('S1A_IW_SLC__1SDV_20211229T231926_20211229T231953_041230_04E66A_3DBE', 'IW2', 'VV', 3)
+    burst_params2 = BurstParams('S1A_IW_SLC__1SDV_20220110T231926_20220110T231953_041405_04EC57_103E', 'IW2', 'VV', 3)
+    metadata = download_bursts([burst_params1, burst_params2])
```
jtherrmann, line 238: Does this just serve as a demo? Also, the `metadata` variable is never used.

src/hyp3_isce2/burst.py
```diff
@@ -0,0 +1,238 @@
...
+class BurstParams:
+    """Class that contains the parameters nessecary to request a burst from the API."""
+
+    granule: str
+    swath: str
+    polarization: str
+    burst_number: int
+
+
+class BurstMetadata:
```
jtherrmann, line 26: I prefer to use meaningful variable names in list comprehensions rather than `x` and `y` for the sake of readability. Not necessarily blocking, but would be nice to have.

jtherrmann, line 26: There's a lot going on in this class and my brain is starting to slow down. This is a self-reminder to come back and look at this more carefully on Monday when I'm refreshed.

src/hyp3_isce2/process.py
```diff
@@ -2,39 +2,71 @@
...
-    return product_file
+    ref_params = BurstParams(reference_scene, f'IW{swath_number}', polarization.upper(), reference_burst_number)
+    sec_params = BurstParams(secondary_scene, f'IW{swath_number}', polarization.upper(), secondary_burst_number)
+    ref_metadata, sec_metadata = download_bursts([ref_params, sec_params])
+
+    is_ascending = ref_metadata.orbit_direction == 'ascending'
+    insar_roi = get_region_of_interest(ref_metadata.footprint, sec_metadata.footprint, is_ascending=is_ascending)
+    dem_roi = ref_metadata.footprint.intersection(sec_metadata.footprint).bounds
+    print(insar_roi, dem_roi)
+    subprocess.run(['python', TOPSAPP, '-h'])
```
jtherrmann, line 50: Hmm, I'm not sure I understand what's going on here. What's the reason for calling `TOPSAPP` with just the `-h` arg?

src/hyp3_isce2/burst.py
forrestfwilliams: Yes this is just a demo. This script has utility outside of the plugin (for creating SAFE-style burst products so I wanted to provide an example. Happy to remove if you don't think it fits here though.

src/hyp3_isce2/process.py
forrestfwilliams: This is a placeholder to prove that we *can* call topsApp in this spot. This will be replaced with the actual processing call once we have the functionality for downloading the input data set up.

src/hyp3_isce2/burst.py
forrestfwilliams: I decided to remove it.

src/hyp3_isce2/burst.py
forrestfwilliams: I've added a new test for this but it's failing 😢. I'll work on fixing!

src/hyp3_isce2/burst.py
```diff
@@ -0,0 +1,339 @@
...
+            point: The geolocation grid point to reformat.
+
+        Returns:
+            A dictionary containing the geolocation grid point's line, pixel, latitude, longitude, and height.
+        """
+        attribs = ['line', 'pixel', 'latitude', 'longitude', 'height']
+        values = {}
+        for attrib in attribs:
+            values[attrib] = float(point.find(attrib).text)
+        return values
```
jtherrmann, line 80:
```suggestion
        return {attrib: float(point.find(attrib).text) for attrib in attribs}
```

src/hyp3_isce2/burst.py
```diff
@@ -0,0 +1,339 @@
...
+    def create_gcp_df(self) -> pd.DataFrame:
+        """Create a dataframe of geolocation grid points.
+
+        Returns:
+            A dataframe containing the geolocation grid points for the burst.
+        """
+        points = self.annotation.findall('.//{*}geolocationGridPoint')
+        gcp_df = pd.DataFrame([self.reformat_gcp(x) for x in points])
+        gcp_df = gcp_df.sort_values(['line', 'pixel']).reset_index(drop=True)
+        return gcp_df
```
jtherrmann, line 91:
```suggestion
        points = self.annotation.findall('.//{*}geolocationGridPoint')
        gcp_df = pd.DataFrame([self.reformat_gcp(point) for point in points])
        return gcp_df.sort_values(['line', 'pixel']).reset_index(drop=True)
```

tests/test_burst.py
```diff
@@ -0,0 +1,91 @@
...
+    assert gcp_df.line.min() == 0
+    assert gcp_df.line.max() == (n_bursts * lines_per_burst) - 1
+
+
+def test_create_geometry():
+    ref_burst = burst.BurstMetadata(load_metadata('reference_descending.xml'), REF_DESC)
+    burst_number = 3
+    real_box = (100.509898817751, 37.69349213923167, 101.5989880944895, 38.00276647361588)
+
+    gcp_df = ref_burst.create_gcp_df()
```
jtherrmann, line 48: In general I would include the literal value for each input to a function that's being tested, rather than call another function to compute the value. But I don't know what the literal value looks like in this case or how cumbersome it would be to include it, so I'll defer to your judgement.

tests/test_burst.py
```diff
@@ -0,0 +1,91 @@
...
+    box = ref_burst.create_geometry(gcp_df)[1]
```
jtherrmann, line 49: So we're grabbing the second value returned by `create_geometry`, but what about the other values returned? Would be nice to `assert` something about those as well.

src/hyp3_isce2/burst.py
```diff
@@ -0,0 +1,339 @@
...
+
+    def create_geometry(self, gcp_df: pd.DataFrame) -> Tuple[geometry.Polygon, Tuple[float], Tuple[float]]:
```
jtherrmann, line 93: OK, this function went over my head, but I think that's more due to my own lack of knowledge in this area rather than any fault of the code. So I left some feedback on the unit test instead.

src/hyp3_isce2/process.py
jtherrmann, line 50:
```suggestion
    # TODO replace with the actual processing call once we have the functionality for downloading the input data
    subprocess.run(['python', TOPSAPP, '-h'])
```

src/hyp3_isce2/burst.py
forrestfwilliams: Fixed!

src/hyp3_isce2/process.py
forrestfwilliams: Thanks for adding it!

tests/test_burst.py
forrestfwilliams: I've added a test for the centroid, but the footprint will be a mess to test, and should be covered by the bounds test.

src/hyp3_isce2/burst.py
```diff
@@ -0,0 +1,335 @@
...
+import requests
+from shapely import geometry
+
+
+URL = 'https://sentinel1-burst.asf.alaska.edu'
+
+
+@dataclass
+class BurstParams:
+    """Class that contains the parameters nessecary to request a burst from the API."""
```
jtherrmann, line 24:
```suggestion
    """Parameters necessary to request a burst from the API."""
```

src/hyp3_isce2/burst.py
```diff
@@ -0,0 +1,335 @@
...
+
+    granule: str
+    swath: str
+    polarization: str
+    burst_number: int
+
+
+class BurstMetadata:
+    """Class that contains the metadata for a burst."""
```
jtherrmann, line 33:
```suggestion
    """Metadata for a burst."""
```

src/hyp3_isce2/burst.py
```diff
@@ -0,0 +1,335 @@
...
+    Returns:
+        The path to the saved SAFE file.
+    """
+    safe_path = base_path / f'{burst.safe_name}.SAFE'
+    annotation_path = safe_path / 'annotation'
+    calibration_path = safe_path / 'annotation' / 'calibration'
+    measurement_path = safe_path / 'measurement'
+    paths = [annotation_path, calibration_path, measurement_path]
+    for p in paths:
+        p.mkdir(parents=True, exist_ok=True)
```
jtherrmann, line 249:
```suggestion
    for path in paths:
        path.mkdir(parents=True, exist_ok=True)
```

src/hyp3_isce2/etc/download_example.py
jtherrmann, line 1: I'm curious how/when/where/by whom this file will be used?

src/hyp3_isce2/etc/download_example.py
```diff
@@ -1,10 +1,19 @@
+"""Miscellaneous examples of downloading data from the ASF Burst
+Exctraction API. These examples are not intended to be run as part of
```
jtherrmann, line 2:
```suggestion
Extraction API. These examples are not intended to be run as part of
```

jtherrmann (approved):

PR #8, opened 2023-04-05
Add ancillary data download functionality

forrestfwilliams:
Adds the ability to download Sentinel-1 orbits and auxiliary calibration files.

The `s1_auxcal` test is slow because it downloads/extracts the data from the ESA archives. Might want to consider adding a `pytest` marker so it is only run during integration testing.

### DIFF
--- a/src/hyp3_isce2/process.py
+++ b/src/hyp3_isce2/process.py
@@ -55,7 +55,7 @@ def topsapp_burst(
 
     download_aux_cal(aux_cal_dir)
     for granule in (ref_params.granule, sec_params.granule):
-        orbit_file, _ = downloadSentinelOrbitFile(granule, orbit_dir)
+        downloadSentinelOrbitFile(granule, orbit_dir)
 
     # TODO replace with the actual processing call once we have the functionality for downloading the input data
     subprocess.run(['python', TOPSAPP, '-h'])
--- a/src/hyp3_isce2/process.py
+++ b/src/hyp3_isce2/process.py
@@ -5,12 +5,21 @@ ISCE2 processing
 import logging
 import os
 import subprocess
-from argparse import ArgumentDefaultsHelpFormatter, ArgumentParser
+from argparse import (
+    ArgumentDefaultsHelpFormatter,
+    ArgumentParser
+)
 from pathlib import Path
 
+from hyp3lib.get_orb import downloadSentinelOrbitFile
+
 from hyp3_isce2 import __version__
-from hyp3_isce2.burst import (BurstParams, download_bursts,
-                              get_region_of_interest)
+from hyp3_isce2.burst import (
+    BurstParams,
+    download_bursts,
+    get_region_of_interest
+)
+
 
 log = logging.getLogger(__name__)
 
@@ -39,6 +48,7 @@ def topsapp_burst(
         azimuth_looks: Number of azimuth looks
         range_looks: Number of range looks
     """
+    orbit_dir = 'orbits'
     ref_params = BurstParams(reference_scene, f'IW{swath_number}', polarization.upper(), reference_burst_number)
     sec_params = BurstParams(secondary_scene, f'IW{swath_number}', polarization.upper(), secondary_burst_number)
     ref_metadata, sec_metadata = download_bursts([ref_params, sec_params])
@@ -47,6 +57,9 @@ def topsapp_burst(
     insar_roi = get_region_of_interest(ref_metadata.footprint, sec_metadata.footprint, is_ascending=is_ascending)
     dem_roi = ref_metadata.footprint.intersection(sec_metadata.footprint).bounds
     print(insar_roi, dem_roi)
+
+    for granule in (ref_params.granule, sec_params.granule):
+        orbit_file, _ = downloadSentinelOrbitFile(granule, orbit_dir)
     # TODO replace with the actual processing call once we have the functionality for downloading the input data
     subprocess.run(['python', TOPSAPP, '-h'])
 
--- a/src/hyp3_isce2/process.py
+++ b/src/hyp3_isce2/process.py
@@ -59,7 +59,7 @@ def topsapp_burst(
     insar_roi = get_region_of_interest(ref_metadata.footprint, sec_metadata.footprint, is_ascending=is_ascending)
     dem_roi = ref_metadata.footprint.intersection(sec_metadata.footprint).bounds
     print(insar_roi, dem_roi)
-    
+
     download_aux_cal(aux_cal_dir)
     for granule in (ref_params.granule, sec_params.granule):
         orbit_file, _ = downloadSentinelOrbitFile(granule, orbit_dir)
--- a/src/hyp3_isce2/process.py
+++ b/src/hyp3_isce2/process.py
@@ -5,20 +5,13 @@ ISCE2 processing
 import logging
 import os
 import subprocess
-from argparse import (
-    ArgumentDefaultsHelpFormatter,
-    ArgumentParser
-)
+from argparse import ArgumentDefaultsHelpFormatter, ArgumentParser
 from pathlib import Path
 
 from hyp3lib.get_orb import downloadSentinelOrbitFile
 
 from hyp3_isce2 import __version__
-from hyp3_isce2.burst import (
-    BurstParams,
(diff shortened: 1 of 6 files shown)


forrestfwilliams:
@jhkennedy or @jtherrmann this is ready for review!

src/hyp3_isce2/process.py
```diff
@@ -5,12 +5,22 @@
 import logging
 import os
 import subprocess
-from argparse import ArgumentDefaultsHelpFormatter, ArgumentParser
+from argparse import (
+    ArgumentDefaultsHelpFormatter,
+    ArgumentParser
+)
```
jhkennedy, line 11: Black did this?

src/hyp3_isce2/s1_auxcal.py
```diff
@@ -0,0 +1,59 @@
...
+    if not isinstance(aux_cal_dir, Path):
+        aux_cal_dir = Path(aux_cal_dir)
+
+    aux_cal_dir.mkdir(exist_ok=True, parents=True)
+    for url in [S1A_AUX_URL, S1B_AUX_URL]:
+        _download_platform(url, aux_cal_dir)
+
+
+if __name__ == '__main__':
+    """Provides a command line interface to download the aux cal files."""
```
jhkennedy, line 58: Does the docstring provide value here over a comment?

tests/conftest.py
```diff
@@ -0,0 +1,9 @@
+import tempfile
+
+import pytest
+
+
+@pytest.fixture()
+def tempdir():
+    with tempfile.TemporaryDirectory() as tempdir:
+        yield tempdir
```
jhkennedy, line 9: pytest has a built-in fixture for this:
https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html

jhkennedy (changes requested):
Overall looks good! I'd prefer a switch to using the `tmp_path` fixture provided by pytest.

src/hyp3_isce2/process.py
forrestfwilliams: No, this was done with [`isort`](https://pycqa.github.io/isort/index.html). I had my line length for this tool set to 88, thus the wrapping.

tests/conftest.py
forrestfwilliams: Didn't know about that, thanks!

src/hyp3_isce2/s1_auxcal.py
forrestfwilliams: nope.

src/hyp3_isce2/process.py
```diff
@@ -8,9 +8,12 @@
 from argparse import ArgumentDefaultsHelpFormatter, ArgumentParser
 from pathlib import Path
 
+from hyp3lib.get_orb import downloadSentinelOrbitFile
```
jtherrmann, line 11: I want to ask @jhkennedy to remind me what is our ultimate goal for https://github.com/ASFHyP3/hyp3-lib. Would we eventually like to eliminate it as a dependency entirely and archive the repo, or are we still fine with using it internally at ASF?

src/hyp3_isce2/process.py
```diff
@@ -47,6 +52,11 @@ def topsapp_burst(
     insar_roi = get_region_of_interest(ref_metadata.footprint, sec_metadata.footprint, is_ascending=is_ascending)
     dem_roi = ref_metadata.footprint.intersection(sec_metadata.footprint).bounds
     print(insar_roi, dem_roi)
+
+    download_aux_cal(aux_cal_dir)
+    for granule in (ref_params.granule, sec_params.granule):
+        orbit_file, _ = downloadSentinelOrbitFile(granule, orbit_dir)
```
jtherrmann, line 58: `orbit_file` is never used, can we just call the function without capturing its return value?

src/hyp3_isce2/s1_auxcal.py
```diff
@@ -0,0 +1,59 @@
...
+from typing import Union
+
+import requests
+
+
+S1A_AUX_URL = 'https://sar-mpc.eu/download/55282da1-679d-4ecf-aeef-d06b024451cf'
+S1B_AUX_URL = 'https://sar-mpc.eu/download/3c8b7c8d-d3de-4381-a19d-7611fb8734b9'
+
+
+def _download_platform(url: str, aux_cal_dir: Path):
```
jtherrmann, line 28: In general I like the use of leading `_` for internal module components, but are we using that convention consistently across this repo? If not, we should probably make that decision.

src/hyp3_isce2/s1_auxcal.py
```diff
@@ -0,0 +1,59 @@
...
+    """Download and extract the aux cal files for a given satellite platform.
+
+    Args:
+        url: URL to download the aux cal files from.
+        aux_cal_dir: Directory to download the aux cal files to.
+    """
+    resp = requests.get(url)
+    resp.raise_for_status()
+
+    content = BytesIO(resp.content)
```
jtherrmann, line 38: I prefer avoiding abbreviations for variable names if possible, but feel free to ignore.
```suggestion
    response = requests.get(url)
    response.raise_for_status()

    content = BytesIO(response.content)
```

src/hyp3_isce2/s1_auxcal.py
```diff
@@ -0,0 +1,59 @@
...
+    """Download and extract the aux cal files for Sentinel-1A/B.
+
+    Args:
+        aux_cal_dir: Directory to download the aux cal files to.
+    """
+    if not isinstance(aux_cal_dir, Path):
+        aux_cal_dir = Path(aux_cal_dir)
+
+    aux_cal_dir.mkdir(exist_ok=True, parents=True)
+    for url in [S1A_AUX_URL, S1B_AUX_URL]:
```
jtherrmann, line 53: Elsewhere in this PR there's a line: `for granule in (ref_params.granule, sec_params.granule)`

It would be nice to be consistent about whether we use `tuple` or `list` in these cases. I used to prefer `tuple` but it seems like most people in the team prefer `list`. But this is a huge nitpick so not blocking at all.

src/hyp3_isce2/s1_auxcal.py
```diff
@@ -0,0 +1,59 @@
...
+        _download_platform(url, aux_cal_dir)
+
+
+if __name__ == '__main__':
+    # Provides a command line interface to download the aux cal files
```
jtherrmann, line 58: I think this is self-explanatory.
```suggestion
```

jtherrmann: Unless we wanted to explain in what cases it would be helpful to run this at the command-line. Is it mainly for development and testing purposes?

src/hyp3_isce2/process.py
forrestfwilliams: Makes sense, I'll remove it.

src/hyp3_isce2/process.py
forrestfwilliams: I think we have long-term plans to either sunset or significantly downsize hyp3-lib to only the useful utilizes but I think we're a ways off from have the time to commit to this effort.

src/hyp3_isce2/s1_auxcal.py
forrestfwilliams: I think we should commit to always using the leading `_` for non-public functions.

src/hyp3_isce2/s1_auxcal.py
forrestfwilliams: Let's use `tuples`.

src/hyp3_isce2/process.py
jtherrmann: Sounds good.

src/hyp3_isce2/s1_auxcal.py
jtherrmann: Sounds good, maybe we should make that change to all our non-public components before merging to `main`.

jtherrmann: I created an issue: https://github.com/ASFHyP3/hyp3-isce2/issues/13

jtherrmann (approved):

PR #29, opened 2023-04-14
Add multi-entrypoint functionality for burst workflow

forrestfwilliams:
* Adds capability to use multiple entrypoints in plugin
* renames burst workflow to `insar_tops_burst`

Will add entrypoint for AVO in a separate PR.


tests/test_entrypoints.py
```diff
@@ -1,3 +1,13 @@
 def test_hyp3_isce2(script_runner):
     ret = script_runner.run('python', '-m', 'hyp3_isce2', '-h')
     assert ret.success
+
+
+def test_insar_tops_burst(script_runner):
+    ret = script_runner.run('python', '-m', 'hyp3_isce2', 'insart_tops_burst', '-h')
```
AndrewPlayer3:
```suggestion
    ret = script_runner.run('python', '-m', 'hyp3_isce2', 'insar_tops_burst', '-h')
```

AndrewPlayer3 (approved):
LGTM!

pyproject.toml
```diff
@@ -42,7 +42,10 @@ Homepage = "https://github.com/ASFHyP3/hyp3-ISCE2"
 Documentation = "https://hyp3-docs.asf.alaska.edu"
 
 [project.scripts]
-topsapp_burst = "hyp3_isce2.process:main"
+insar_tops_burst = "hyp3_isce2.insar_tops_burst:main"
```
jhkennedy, line 45: Overall, I still don't really understand the motivation for this entry point -- it looks like it's just o make it so you can run:
```shell
insar_tops_burst [ARGS]
```
instead of 
```shell
python -m hyp3_isce2 [ARGS]
```
as they only differ by the `--bucket` and `--bucket-prefix` arguments, which *are* optional. 

If it's a convenience thing (not wanting to type `python -m ` in front) then I'd ranter see a script entrypoint like:
```suggestion
hyp3_isce2 = "hyp3_isce2.__main__:main"
```
you you'd be able to just run:
```shell
hyp3_isce2 [ARGS]
```

**If** it's for actually different behavior anticipated in the future like some of our other "science" code, I can see adding an extra script entrypoint for convenience, but I'd rather do it *then*.

CHANGELOG.md
```diff
@@ -6,6 +6,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/)
 and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.2.0]
+### Added
+* Multi-entrypoint capabilities to plugin
```
jhkennedy, line 11:
```suggestion
* It's now possible to register multiple [HyP3 entry points (workflows) and run them through the main hyp3_isce2 entry point. 
```
Maybe something like this that's a little more descriptive. We should probably add a "how to" for adding a new hyp3 entry point, either in the README or `docs`.

CHANGELOG.md
```diff
@@ -6,6 +6,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+
+### Changed
+* Burst workflow name to `insar_tops_burst`
```
jhkennedy, line 14:
```suggestion
* Burst workflow renamed to `insar_tops_burst` from `topsapp_burst`
```

jhkennedy: Although I am suggesting we drop the burst specific console script entry point and (maybe) provide a `hyp3_isce2` one.

src/hyp3_isce2/insar_tops_burst.py
```diff
@@ -85,23 +92,89 @@ def topsapp_burst(
...
+
 
+def main():
+    """Main entrypoint for the burst workflow"""
+    parser = _get_cli(interface='main')
+    args = parser.parse_args()
 
-if __name__ == "__main__":
-    main()
+    run_insar_tops_burst(args)
```
jhkennedy, line 180: Dropping the burst specific entry point would me all this could be condensed back into `main`.

pyproject.toml
jhkennedy: :thinking: I suppose, for the non-default HyP3 entry point, you'll need to add `++process [PROC_NAME]`  which does get a bit long:
```
python -m hyp3_isce2 ++process [PROC_NAME] [ARGS]
# Or
hyp3_isce2 ++process [PROC_NAME] [ARGS]
```
so I guess having specific workflow entry points *might* make sense. In that case, 

I think I'd rather just see the `[project.scripts]` and `[project.entry-points.hyp3]` sections of `pyproject.toml` be exactly the same, or potentially better, not using the argparse trick at all and building out a proper CLI with sub parsers:
```shell
hyp3_isce2 [PROC_NAME] [ARGS]
```

PR #30, opened 2023-04-17
Fix 0-index bug in burst.py

forrestfwilliams:
In an earlier update, we changed to using zero-indexed burst numbers but did not switch over to this paradigm in the `create_geometry` method.


jhkennedy (approved):

forrestfwilliams:
I've done some end-to-end test to make sure that the burst indexing is correct.

PR #32, opened 2023-04-18
Update Readme

forrestfwilliams:


asjohnston-asf:
I'd prefer to see Usage moved first ahead of Background and Developer Setup. I suspect the `python` and `docker run` examples are the most useful part of the document, and I suspect that's what I'll be looking for half the time I come to the readme, so I'd like to see them as early as possible.

jhkennedy (approved):
looks good, just one small thing -- approval would stand with it

Issue #20, opened 2023-04-11
Support Python 3.10


jtherrmann:
When I specify `python=3.10` in the environment file, I get the following conflicts:

[python-3-10-conflicts-hyp3-isce2.txt](https://github.com/ASFHyP3/hyp3-isce2/files/11195071/python-3-10-conflicts-hyp3-isce2.txt)

I have not yet inspected this output carefully enough to identify the root of the conflicts. For now, it seems we cannot support Python 3.10.

PR #33, opened 2023-04-19
Upgrade to ISCE2 2.6.3

forrestfwilliams:
Fixes `numpy` incompatibility and **removes annoying Stanford license printout!!!**

I have confirmed that all tests pass in new environment build, and interferograms are produced as expected.


environment.yml
```diff
@@ -4,10 +4,10 @@ channels:
   - nodefaults
 dependencies:
   # python and numpy pins: https://github.com/ASFHyP3/hyp3-isce2/issues/14
```
jtherrmann, line 6:
```suggestion
```

jtherrmann:
This might also resolve https://github.com/ASFHyP3/hyp3-isce2/issues/20 (we will have to see if there are still conflicts when installing Python 3.10).

PR #34, opened 2023-04-19
Release v0.2.0

forrestfwilliams:


CHANGELOG.md
jtherrmann, line 1: These can all be combined into a single `0.2.0` entry since we haven't actually released since `0.1.0`.

jtherrmann (changes requested):

CHANGELOG.md
forrestfwilliams: @jtherrmann I created this [PR](https://github.com/ASFHyP3/hyp3-isce2/pull/35) to fix the changelog.

jtherrmann: @forrestfwilliams I reviewed that PR.

CHANGELOG.md
```diff
@@ -6,6 +6,19 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+* It's now possible to register multiple HyP3 entry points (workflows) and run them through the main hyp3_isce2 entry point. 
+* A new entrypoint and skeleton process file for a stripmap process.
+* Concurrent download functionality for burst extractor calls.
+
+### Changed
+* Burst workflow renamed to `insar_tops_burst` from `topsapp_burst`.
+
+### Fixed
+* A zero-index bug in `burst.py` that led to incorrect geolocation of products.
+* Typo in the `release-checklist-comment` workflow.
```
jtherrmann, line 20: I *think* we normally don't include newlines between the subsections, but it shouldn't cause any problems? We could fix it later with a bumpless PR into `develop`.

jhkennedy: @jtherrmann it's not actually standard. *I* prefer the new lines because it's more readable, so most of the plugins do. James and @asjohnston-asf preferred compact, so Hyp3 and other platform things typically do not. 

Would be a good idea to decide and standardize.

jtherrmann (approved):

CHANGELOG.md
jtherrmann: @jhkennedy Gotcha, thanks for clarifying.

jhkennedy: @jtherrmann to standardize, we might be able to use this:
https://github.com/DavidAnson/markdownlint

or find a GitHub Action which uses it.

jhkennedy: Moving the markdown linting discussion to here: https://github.com/ASFHyP3/actions/issues/65

PR #36, opened 2023-04-21
Fix bug in creation of burst bounding boxes

forrestfwilliams:
Running `insar_tops_burst` with the following parameters revealed that bounding boxes for burst were note being created correctly. Rather than determine what edge case we were running into, I opted to switch over to using ISCE2's methodology for creating bounding boxes directly. This will ensure that our bounding boxes always agree with the methodology used by `topsApp`

This PR also fixes a bug in `test_burst.py`. We (I) accidentally had two descending test cases for `test_get_region_of_interest`, not a ascending AND descending test case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ## [0.2.1]
 ### Fixed
 * Inaccurate generation of burst bounding boxes in a Qinghai China example pair
+* `test_get_region_of_interest` so that it has an ascending and descending test case, not two descending. This required changing the `reference_ascending.xml` and `secondary_ascending.xml` files in `tests/data`.
 
 ### Added
 * The `get_isce2_burst_bbox` function to calculate burst bounding boxes using ISCE2 directly
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,17 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/)
 and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+
+## [0.2.1]
+### Fixed
+* Inaccurate generation of burst bounding boxes in a Qinghai China example pair
+
+### Added
+* The `get_isce2_burst_bbox` function to calculate burst bounding boxes using ISCE2 directly
+
+### Removed
+* The `reformat_gcp`, `create_gcp_df`, and `create_geometry` methods from the `BurstMetadata` class because they are superseded by `get_isce2_burst_bbox`
+
 ## [0.2.0]
 ### Added
 * It's now possible to register multiple HyP3 entry points (workflows) and run them through the main hyp3_isce2 entry point. 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,15 +8,14 @@ and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 
 ## [0.2.1]
-### Fixed
-* Inaccurate generation of burst bounding boxes in a Qinghai China example pair
-* `test_get_region_of_interest` so that it has an ascending and descending test case, not two descending. This required changing the `reference_ascending.xml` and `secondary_ascending.xml` files in `tests/data`.
-
 ### Added
-* The `get_isce2_burst_bbox` function to calculate burst bounding boxes using ISCE2 directly
+* The `get_isce2_burst_bbox` function to calculate burst bounding boxes using ISCE2 directly to fix inaccurate generation of burst bounding boxes.
 
 ### Removed
 * The `reformat_gcp`, `create_gcp_df`, and `create_geometry` methods from the `BurstMetadata` class because they are superseded by `get_isce2_burst_bbox`
+
+### Fixed
+* `test_get_region_of_interest` so that it has an ascending and descending test case, not two descending. This required changing the `reference_ascending.xml` and `secondary_ascending.xml` files in `tests/data`.
 
 ## [0.2.0]
 ### Added
--- a/src/hyp3_isce2/burst.py
+++ b/src/hyp3_isce2/burst.py
@@ -215,10 +215,10 @@ def spoof_safe(burst: BurstMetadata, burst_tiff_path: Path, base_path: Path = Pa
     return safe_path
 
 
-def get_isce2_burst_bbox(params: BurstParams) -> geometry.Polygon:
+def get_isce2_burst_bbox(params: BurstParams, base_dir: Path = Path.cwd()) -> geometry.Polygon:
     s1_obj = Sentinel1()
     s1_obj.configure()
-    s1_obj.safe = [f'{params.granule}.SAFE']
+    s1_obj.safe = [str(base_dir / f'{params.granule}.SAFE')]
     s1_obj.swathNumber = int(params.swath[-1])
     s1_obj.parse()
     snwe = s1_obj.product.bursts[params.burst_number].getBbox()
--- a/src/hyp3_isce2/burst.py
+++ b/src/hyp3_isce2/burst.py
@@ -7,7 +7,7 @@ from dataclasses import dataclass
 from pathlib import Path
 from typing import Iterator, List, Tuple, Union
 
-import isce  # noqa
+import isce  # noqa: F401
 import requests
 from isceobj.Sensor.TOPS.Sentinel1 import Sentinel1
 from lxml import etree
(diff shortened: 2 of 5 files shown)


forrestfwilliams:
Here's an example of a command that will fail because ISCE2 finds too many bursts that overlap with the region of interest we generate:

```bash
insar_tops_burst \
  --reference-scene S1A_IW_SLC__1SDV_20200604T022251_20200604T022318_032861_03CE65_7C85 \
  --secondary-scene S1A_IW_SLC__1SDV_20200616T022252_20200616T022319_033036_03D3A3_5D11 \
  --swath-number 2 \
  --polarization VV \
  --reference-burst-number 7 \
  --secondary-burst-number 7 \
  --azimuth-looks 4 \
  --range-looks 20
  ```

**Comment by forrestfwilliams** [comment]
Trufflehog is failing because we updated the test xml files.

File: CHANGELOG.md [hunk 1]
```diff
@@ -6,6 +6,18 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/)
 and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+
+## [0.2.1]
+### Fixed
+* Inaccurate generation of burst bounding boxes in a Qinghai China example pair
```
jtherrmann, line 12: I'm not sure the Changelog is the best place to call out the particular example that failed. Is this the pair used in the command that you commented [here](https://github.com/ASFHyP3/hyp3-isce2/pull/36#issuecomment-1518208625)? I think that comment is probably enough (or you could document it in an issue if you want to make it more visible).

CHANGELOG.md
```diff
@@ -6,6 +6,18 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/)
 and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+
+## [0.2.1]
+### Fixed
```
jtherrmann, line 11: I think we want to move `Fixed` to be the last section so that we follow the order from https://keepachangelog.com/en/1.0.0/: Added, Removed, Fixed

CHANGELOG.md
jtherrmann: Also, is this part of the same change as the bullet under `Added`? We could just combine these into a single bullet.

src/hyp3_isce2/burst.py
```diff
@@ -7,8 +7,9 @@
 from pathlib import Path
 from typing import Iterator, List, Tuple, Union
 
-import pandas as pd
+import isce  # noqa
```
jtherrmann, line 10: Why the `# noqa` here?

src/hyp3_isce2/burst.py
```diff
@@ -262,7 +214,19 @@ def spoof_safe(burst: BurstMetadata, burst_tiff_path: Path, base_path: Path = Pa
...
-def get_region_of_interest(poly1: geometry.Polygon, poly2: geometry.Polygon, is_ascending: bool = True) -> Tuple[float]:
+def get_isce2_burst_bbox(params: BurstParams, base_dir: Path = Path.cwd()) -> geometry.Polygon:
+    s1_obj = Sentinel1()
+    s1_obj.configure()
+    s1_obj.safe = [str(base_dir / f'{params.granule}.SAFE')]
+    s1_obj.swathNumber = int(params.swath[-1])
+    s1_obj.parse()
+    snwe = s1_obj.product.bursts[params.burst_number].getBbox()
+    # convert from south, north, west, east -> minx, miny, maxx, maxy
+    bbox = geometry.box(snwe[2], snwe[0], snwe[3], snwe[1])
```
jtherrmann, line 225: For some reason I thought South was generally equivalent to max Y? We should talk about this to clear up my confusion.

src/hyp3_isce2/burst.py
```diff
@@ -262,7 +214,19 @@ def spoof_safe(burst: BurstMetadata, burst_tiff_path: Path, base_path: Path = Pa
...
+    return bbox
+
+
+def get_region_of_interest(bbox1: geometry.Polygon, bbox2: geometry.Polygon, is_ascending: bool = True) -> Tuple[float]:
```
jtherrmann, line 229: The docstring argument names need to be updated (but I can't add a suggestion for lines that haven't changed). Also maybe worth specifying the order of the bounds?

CHANGELOG.md
jtherrmann: Oh sorry, now I'm realizing this probably refers to the changes in [`download_example.py`](https://github.com/ASFHyP3/hyp3-isce2/pull/36/files#diff-1c2b82e1309d796ea81d82e401a27118fc3e3dfd564fe269e989f783dce2fa35)?

tests/test_burst.py
```diff
@@ -65,27 +39,31 @@ def test_spoof_safe(tmp_path, mocker, pattern):
     assert len(list(tmp_path.glob(pattern))) == 1
 
 
-@pytest.mark.parametrize('orbit', ('ascending', 'descending'))
-def test_get_region_of_interest(orbit):
-    options = {'descending': REF_DESC, 'ascending': REF_ASC}
+@pytest.mark.parametrize('orbit', ('ascending', 'ascending'))
```
jtherrmann, line 42: Shouldn't one of these still be `descending`?

```suggestion
@pytest.mark.parametrize('orbit', ('ascending', 'descending'))
```

tests/test_burst.py
```diff
@@ -65,27 +39,31 @@ def test_spoof_safe(tmp_path, mocker, pattern):
...
+def test_get_region_of_interest(tmp_path, orbit):
```
jtherrmann, line 43: Could we get a docstring for this function that just describes in high-level language what exactly we're testing with the three assertions?

CHANGELOG.md
forrestfwilliams: Nope, your second comment was correct. I've updated the changelog.

CHANGELOG.md
forrestfwilliams: done

src/hyp3_isce2/burst.py
forrestfwilliams: To import `isceobj`, you must first `import isce`, even if you're not directly using `isce`. Flake8 flags this as an unused import unless specify `# noqa`.

src/hyp3_isce2/burst.py
forrestfwilliams: No worries, South should be `miny`. When I get confused, or just want to check something, I go to vertex and move my cursor around the map. It displays your latitude/longitude in the upper left corner so you can see how these values change as you move.

src/hyp3_isce2/burst.py
forrestfwilliams: Ah, forgot to change this. Thanks!

tests/test_burst.py
forrestfwilliams: will do!

src/hyp3_isce2/burst.py
jtherrmann, line 10:
```suggestion
import isce  # noqa: F401
```

tests/test_burst.py
```diff
@@ -66,26 +40,47 @@ def test_spoof_safe(tmp_path, mocker, pattern):
...
+          +---------------+
+          |               |
+       +--+------------+  |
+       |  |            |  |
+    +--+--+---------+--+--+
+    |  |            |  |
+    |  +------------+--0
+    |               |
+    +---------------+
+    The diagram for an ascending orbit is the same, but rotated 180 degrees.
```
jtherrmann, line 59: Not sure I fully understand what's going on here, but I really like this diagram regardless.

jtherrmann (approved):
I left one minor suggestion (the `noqa` comment) that would be good to commit before merging.

tests/test_burst.py
forrestfwilliams: Yeah, it's a bit strange!

jtherrmann (approved):

src/hyp3_isce2/burst.py
```diff
@@ -262,23 +214,49 @@ def spoof_safe(burst: BurstMetadata, burst_tiff_path: Path, base_path: Path = Pa
...
 
 
-def get_region_of_interest(poly1: geometry.Polygon, poly2: geometry.Polygon, is_ascending: bool = True) -> Tuple[float]:
+def get_isce2_burst_bbox(params: BurstParams, base_dir: Optional[Path] = None) -> geometry.Polygon:
+    """Get the bounding box of a Sentinel-1 burst using ISCE2.
+    Using ISCE2 directly ensures that the bounding box is the same as the one used by ISCE2 for processing.
+
+    args:
+        params: The burst parameters.
+        base_dir: The directory containing the SAFE file.
```
jhkennedy, line 223: Might want to add what happens here if None is provided

jtherrmann (approved):

jtherrmann (approved):

PR #37, opened 2023-04-25
Create a new InSAR TOPS workflow

forrestfwilliams:
This PR enables TOPS processing for full SLCs


src/hyp3_isce2/insar_tops.py
```diff
@@ -0,0 +1,120 @@
...
+
+def insar_tops(
+    reference_scene: str,
+    secondary_scene: str,
+    swaths: list = [1, 2, 3],
+    polarization: str = 'VV',
+    azimuth_looks: int = 4,
+    range_looks: int = 20,
+) -> Path:
+    """Create a burst interferogram
```
AndrewPlayer3, line 38: Should probably be "Create an interferogram" since this is a full SLC interferogram.

src/hyp3_isce2/slc.py
```diff
@@ -0,0 +1,71 @@
...
+        bbox = geometry.box(*footprint.bounds)
+        bboxs.append(bbox)
+
+    intersection = bboxs[0].intersection(bboxs[1])
+    return intersection.bounds
+
+
+if __name__ == '__main__':
+    ref_granule = 'S1A_IW_SLC__1SDV_20200604T022251_20200604T022318_032861_03CE65_7C85'
+    sec_granule = 'S1A_IW_SLC__1SDV_20200616T022252_20200616T022319_033036_03D3A3_5D11'
```
AndrewPlayer3, line 66: What are these hardcoded granule names for?

tests/conftest.py
```diff
@@ -0,0 +1,10 @@
+import os
+from pathlib import Path
+
+import pytest
+
+
+@pytest.fixture()
+def test_data_dir():
+    here = Path(os.path.dirname(__file__))
+    return here / 'data'
```
AndrewPlayer3, line 10: Should this assert something? Like whether the data dir exists, rather than returning it.

src/hyp3_isce2/insar_tops.py
forrestfwilliams: Thanks for the catch!

src/hyp3_isce2/slc.py
forrestfwilliams: I went ahead and removed the whole code block, since we don't need it.

AndrewPlayer3 (approved):
Looks good to me, now!

PR #38, opened 2023-04-25
Update release.yml

jhkennedy:
Since this is our plugin, we can use the defaults here. Also, this switches back to using the `tools-bot` label as we do for all our other repos and because we're using the Tools Bot access key.


jhkennedy:
Trufflehog is just grumpy about XML changes; safe to ignore

AndrewPlayer3 (approved):
LGTM - identical to the way it is in our other repos.

PR #44, opened 2023-05-15
Add ability to pass EarthData credentials as environment variables

forrestfwilliams:
This capability is the preferred way to pass the user's credentials to the docker container. I've checked this functionality by building a local version of this container `docker build -t hyp3-isce2:latest .` using the `credentials` branch then running:

```bash
docker run -it --rm \
    -e EARTHDATA_USERNAME={YOUR_USERNAME_HERE} \
    -e EARTHDATA_PASSWORD={YOUR_PASSWORD_HERE} \
    hyp3-isce2:latest \
    ++process insar_tops_burst \
    --reference-scene S1A_IW_SLC__1SDV_20200604T022251_20200604T022318_032861_03CE65_7C85 \
    --secondary-scene S1A_IW_SLC__1SDV_20200616T022252_20200616T022319_033036_03D3A3_5D11 \
    --swath-number 2 \
    --polarization VV \
    --reference-burst-number 7 \
    --secondary-burst-number 7 \
    --azimuth-looks 4 --range-looks 20
```

This worked as expected.

This PR does not create functionality to pass credentials as arguments to the workflows. I'd prefer to not add this functionality and instead limit credential passing to either a `.netrc` or as environment variables.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ## [0.3.0]
 ### Added
 * `insar_tops` workflow for processing of full Sentinel-1 SLCs.
-* Ability to pass EarthData username and password as environment variables to workflows. This allows the credentials to be passed to the docker container via a `-e` command.
+* Ability to pass EarthData username and password as environment variables to workflows. This allows the credentials to be passed to the Docker container via the `-e` option.
 
 ## [0.2.1]
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/)
 and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.2.2]
+### Added
+* Ability to pass EarthData username and password as environment variables to workflows. This allows the credentials to be passed to the docker container via a `-e` command.
 
 ## [0.2.1]
 ### Added
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ python -m hyp3_isce2 ++process insar_tops_burst \
 This command will create a Sentinel-1 interferogram that contains a deformation signal related to a 2020 Iranian earthquake.
 To learn about the arguments for each workflow, look at the help documentation (`python -m hyp3_isce2 ++process [WORKFLOW_NAME] --help`).
 
-For all workflows the user will need to provide their EarthData login credentials to download input data. If you do not already have an account, you can sign up [here](https://urs.earthdata.nasa.gov/home). Your credentials can either be passed to the workflows via environment variables (`EARTHDATA_USERNAME` and `EARTHDATA_PASSWORD`), or via your `.netrc` file. If you haven't set up a `.netrc` file before, check out this [guide](https://harmony.earthdata.nasa.gov/docs#getting-started) to get started.
+For all workflows the user will need to provide their Earthdata login credentials to download input data. If you do not already have an account, you can sign up [here](https://urs.earthdata.nasa.gov/home). Your credentials can either be passed to the workflows via environment variables (`EARTHDATA_USERNAME` and `EARTHDATA_PASSWORD`), or via your `.netrc` file. If you haven't set up a `.netrc` file before, check out this [guide](https://harmony.earthdata.nasa.gov/docs#getting-started) to get started.
 
 The ultimate goal of this project is to create a docker container that can run ISCE2 workflows within a HyP3 deployment. To run the current version of the project's container, use this command:
 ```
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/)
 and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.2.2]
+### Added
+* Ability to pass EarthData username and password as environment variables to workflows. This allows the credentials to be passed to the docker container via a `-e` command.
```
jtherrmann, line 11:
```suggestion
* Ability to pass EarthData username and password as environment variables to workflows. This allows the credentials to be passed to the Docker container via the `-e` option.
```

src/hyp3_isce2/__main__.py
```diff
@@ -14,15 +17,19 @@ def main():
...
-            'insar_stripmap'
-        ],
+        choices=['insar_tops_burst', 'insar_stripmap'],
         default='insar_tops_burst',
         help='Select the HyP3 entrypoint to use',  # HyP3 entrypoints are specified in `pyproject.toml`
     )
 
     args, unknowns = parser.parse_known_args()
+
+    # NOTE: Earthdata username and password should be provided as environment variables, or be present in your .netrc
```
jtherrmann, line 27: This comment should probably be moved to the README for better visibility to users.

jtherrmann: Or perhaps it's not needed at all, if users always pass these to the Docker container?

CHANGELOG.md
forrestfwilliams: Changed!

src/hyp3_isce2/__main__.py
forrestfwilliams: I added a note to the readme, since it's worth covering this up-front.

README.md
```diff
@@ -25,9 +26,16 @@ python -m hyp3_isce2 ++process insar_tops_burst \
 This command will create a Sentinel-1 interferogram that contains a deformation signal related to a 2020 Iranian earthquake.
 To learn about the arguments for each workflow, look at the help documentation (`python -m hyp3_isce2 ++process [WORKFLOW_NAME] --help`).
 
+For all workflows the user will need to provide their EarthData login credentials to download input data. If you do not already have an account, you can sign up [here](https://urs.earthdata.nasa.gov/home). Your credentials can either be passed to the workflows via environment variables (`EARTHDATA_USERNAME` and `EARTHDATA_PASSWORD`), or via your `.netrc` file. If you haven't set up a `.netrc` file before, check out this [guide](https://harmony.earthdata.nasa.gov/docs#getting-started) to get started.
```
jtherrmann, line 29:
```suggestion
For all workflows the user will need to provide their Earthdata login credentials to download input data. If you do not already have an account, you can sign up [here](https://urs.earthdata.nasa.gov/home). Your credentials can either be passed to the workflows via environment variables (`EARTHDATA_USERNAME` and `EARTHDATA_PASSWORD`), or via your `.netrc` file. If you haven't set up a `.netrc` file before, check out this [guide](https://harmony.earthdata.nasa.gov/docs#getting-started) to get started.
```

src/hyp3_isce2/__main__.py
```diff
@@ -24,6 +27,12 @@ def main():
     )
 
     args, unknowns = parser.parse_known_args()
+
+    username = os.getenv('EARTHDATA_USERNAME')
+    password = os.getenv('EARTHDATA_PASSWORD')
+    if username and password:
+        write_credentials_to_netrc_file(username, password, append=False)
+
```
jtherrmann, line 35: Does an error occur somewhere if these are not provided? Should we raise an error explicitly as an `else` here?

jtherrmann (approved):

src/hyp3_isce2/__main__.py
```diff
@@ -33,6 +30,9 @@ def main():
     if username and password:
         write_credentials_to_netrc_file(username, password, append=False)
```
jtherrmann, line 31: Ah, now that I think about it, do we want to overwrite the user's netrc file without their explicit permission? I assume most users would *probably* be supplying the same Earthdata credentials already present in their netrc file, but not sure we should make that assumption for them.

jtherrmann: Oh, nevermind, I didn't realize this was the hyp3 entrypoint, so it should be safe to assume that it's running in AWS?

jtherrmann: But if we do intend users to ever run this entrypoint locally, we should probably not automatically overwrite their netrc file.

forrestfwilliams: with `append=False`, this function will only not append to the user's netrc file, it will only create it if it doesn't exist.  See function definition [here](https://github.com/ASFHyP3/hyp3-lib/blob/b0e160d1a3c8ef42eb20a3c3cff5f6d5faedee1d/hyp3lib/fetch.py#L16).

jtherrmann: Ah! For some reason I thought that meant "overwrite the user's netrc file instead of appending to it".

PR #45, opened 2023-05-16
First draft of product packaging

asjohnston-asf:
- [x] include desired output files
  - [x] wrapped phase geotiff
  - [x] unwrapped phase geotiff
  - [x] connected components geotiff
  - [x] coherence (correlation) geotiff
  - [x] DEM geotiff
  - [x] parameter file
- [ ] populate parameter file
- [x] implement naming scheme
- [x] tiling/compression for output geotiffs
- [x] UTM projection for output geotiffs
- [x] nodata values for output geotiffs
- [x] consistent phase value convention with hyp3-gamma products
- [x] changelog entry

stretch goals:
- [x] unwrapped phase browse image
- [ ] upload unzipped output files to s3
- [x] ++omp-num-threads parameter
- [ ] ensure common pixel grid for geotiffs within a single job
- [ ] ensure common pixel grid for geotiffs across multiple jobs

PR #46, opened 2023-05-16
set nodata for output tif files

cirrusasf:


src/hyp3_isce2/insar_tops_burst.py
```diff
@@ -175,23 +175,27 @@ def main():
...
     gdal.Translate(
         destName=f'{product_name}/{product_name}_corr.tif',
         srcDS=str(product_dir / 'phsig.cor.geo'),
+        noData=0,
         creationOptions=['TILED=YES', 'COMPRESS=LZW', 'NUM_THREADS=ALL_CPUS'],
     )
     gdal.Translate(
         destName=f'{product_name}/{product_name}_conn_comp.tif',
         srcDS=str(product_dir / 'filt_topophase.unw.conncomp.geo'),
+        noData=-1,
```
asjohnston-asf, line 190: gdal emits `Warning 1: for band 1, nodata value has been clamped to 0, the original value being out of range.` here, and the output geotiff ends up with noData=0
```
Band 1 Block=256x256 Type=Byte, ColorInterp=Gray
  NoData Value=0
```

The padding for the connected components geotiff has a data value of 0, but there are also some zero value pixels in the interior of the scene. I suspect those interior zeros represent valid data. @forrestfwilliams any thoughts?
Screenshot from 2023-05-16 11-13-16

asjohnston-asf: I don't see any evidence that the ISCE2 outputs in the `merged/` directory have noData values set. We're effectively masking out the padding value for all four output geotiffs. I think this is the breakdown:


| tiff | padding value | data min | data max |
| --- | --- | --- | --- |
| wrapped phase | 0 | -pi | pi |
| unwrapped phase | 0 | -inf (?) | inf (?) |
| connected components | 0 | 0 (?) | 256 |
| coherence | 0 | 0 | 1 |

forrestfwilliams: @asjohnston-asf to the connected components question, yes the interior zeroes are NoData. All the connected components pixels with the same value should be contiguous. The only exception is `0` e.g., the NoData value.

forrestfwilliams: Here's how I'd update your table with the connected components info and the fact that coherence will never reach the value of zero:

| tiff | padding value | data min | data max |
| --- | --- | --- | --- |
| wrapped phase | 0 | -pi | pi |
| unwrapped phase | 0 | -inf (?) | inf (?) |
| connected components | 0 | 1 | 256 |
| coherence | 0 | >0 | 1 |

asjohnston-asf: That checks out, I'd forgotten valid connected components are contiguous. That alleviate my concerns of slapping nodata=0 on connected components are coherence. I'll keep mulling if there's a better approach to the phase tifs.

forrestfwilliams:
Closing since these changes were incorporated in PR #45.

PR #47, opened 2023-05-17
Forrest Packaging Updates

forrestfwilliams:
1. Move translation to separate function
2. Create function for determining utm projection
3. Add nodata setting
4. Warp outputs to an appropriate UTM projection

### DIFF
--- a/src/hyp3_isce2/insar_tops_burst.py
+++ b/src/hyp3_isce2/insar_tops_burst.py
@@ -126,24 +126,26 @@ def make_parameter_file(out_path: Path, reference_scene: str, secondary_scene: s
 
 
 def translate_outputs(product_dir: Path, product_name: str):
-    # TODO need to set nodata values
-    # TODO what output projection do we want? currently EPSG:4326
-    utm_crs = get_utm_proj(product_dir / 'filt_topophase.unw.geo')
-
     gdal.Translate(
         destName=f'{product_name}/{product_name}_unw_phase.tif',
         srcDS=str(product_dir / 'filt_topophase.unw.geo'),
         bandList=[2],
+        format='GTiff',
+        noData=0,
         creationOptions=['TILED=YES', 'COMPRESS=LZW', 'NUM_THREADS=ALL_CPUS'],
     )
     gdal.Translate(
         destName=f'{product_name}/{product_name}_corr.tif',
         srcDS=str(product_dir / 'phsig.cor.geo'),
+        format='GTiff',
+        noData=0,
         creationOptions=['TILED=YES', 'COMPRESS=LZW', 'NUM_THREADS=ALL_CPUS'],
     )
     gdal.Translate(
         destName=f'{product_name}/{product_name}_conn_comp.tif',
         srcDS=str(product_dir / 'filt_topophase.unw.conncomp.geo'),
+        format='GTiff',
+        noData=0,
         creationOptions=['TILED=YES', 'COMPRESS=LZW', 'NUM_THREADS=ALL_CPUS'],
     )
     subprocess.call(
@@ -165,11 +167,19 @@ def translate_outputs(product_dir: Path, product_name: str):
             'COMPRESS=LZW',
             '--creation-option',
             'NUM_THREADS=ALL_CPUS',
+            '--NoDataValue=0',
         ]
     )
 
-    for file in [str(x) for x in Path(product_dir).glob('*.tif')]:
-        gdal.Warp(file, file, dstSRS=utm_crs)
+    utm_crs = get_utm_proj(product_dir / 'filt_topophase.unw.geo')
+    files = [str(x) for x in Path(product_name).glob('*.tif')]
+    for file in files:
+        gdal.Warp(
+            file,
+            file,
+            dstSRS=f'epsg:{utm_crs.to_epsg()}',
+            creationOptions=['TILED=YES', 'COMPRESS=LZW', 'NUM_THREADS=ALL_CPUS'],
+        )
 
 
 def main():
@@ -204,7 +214,7 @@ def main():
     )
 
     log.info('ISCE2 TopsApp run completed successfully')
-
+    product_dir = Path.home() / 'data' / 'example_job'
     product_name = get_product_name(
         args.reference_scene,
         args.secondary_scene,
@@ -213,8 +223,8 @@ def main():
         args.swath_number,
         args.polarization,
     )
-    os.mkdir(product_name)
-    translate_outputs(product_dir, product_name)
+    Path(product_name).mkdir(exist_ok=True)
+    translate_outputs(product_dir / 'merged', product_name)
     make_parameter_file(
         Path(f'{product_name}/{product_name}.json'),
         args.reference_scene,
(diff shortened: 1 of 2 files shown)


src/hyp3_isce2/insar_tops_burst.py
```diff
@@ -126,6 +118,64 @@ def make_parameter_file(
...
+        product_dir: Path to the ISCE merge directory
+        product_name: Name of the product
+    """
+    ISCE2_DATASET = namedtuple('ISCE2_DATASET', ['name', 'suffix', 'band'])
+    datasets = [
+        ISCE2_DATASET('filt_topophase.unw.geo', 'unw_phase', 2),
+        ISCE2_DATASET('phsig.cor.geo', 'corr', 1),
+        ISCE2_DATASET('z.rdr.full.geo', 'dem', 1),
+        ISCE2_DATASET('filt_topophase.unw.conncomp.geo', 'conncomp', 1),
+        ISCE2_DATASET('filt_topophase.flat.geo', 'wrapped_phase', 1),
```
jtherrmann, line 134: Camel case is the convention for class names.
```suggestion
    ISCE2Dataset = namedtuple('ISCE2_DATASET', ['name', 'suffix', 'band'])
    datasets = [
        ISCE2Dataset('filt_topophase.unw.geo', 'unw_phase', 2),
        ISCE2Dataset('phsig.cor.geo', 'corr', 1),
        ISCE2Dataset('z.rdr.full.geo', 'dem', 1),
        ISCE2Dataset('filt_topophase.unw.conncomp.geo', 'conncomp', 1),
        ISCE2Dataset('filt_topophase.flat.geo', 'wrapped_phase', 1),
```

src/hyp3_isce2/insar_tops_burst.py
```diff
@@ -126,6 +118,64 @@ def make_parameter_file(
...
+        in_file = str(product_dir / dataset.name)
+
+        if dataset.suffix == 'wrapped_phase':
+            cmd = (
+                f'gdal_calc.py --outfile {out_file} -A {in_file} '
+                '--calc angle(A) --type Float32 --format GTiff --NoDataValue=0 '
+                '--creation-option TILED=YES --creation-option COMPRESS=LZW --creation-option NUM_THREADS=ALL_CPUS'
+            )
+
+            subprocess.call(cmd.split(' '))
```
jtherrmann, line 148: If there's no Python binding available, then should probably use [`check_call`](https://docs.python.org/3/library/subprocess.html#subprocess.check_call) so that an exception is raised if appropriate.
```suggestion
            cmd = (
                f'gdal_calc.py --outfile {out_file} -A {in_file} '
                '--calc angle(A) --type Float32 --format GTiff --NoDataValue=0 '
                '--creation-option TILED=YES --creation-option COMPRESS=LZW --creation-option NUM_THREADS=ALL_CPUS'
            )

            subprocess.check_call(cmd.split(' '))
```

src/hyp3_isce2/insar_tops_burst.py
```diff
@@ -126,6 +118,64 @@ def make_parameter_file(
...
+    ISCE2_DATASET = namedtuple('ISCE2_DATASET', ['name', 'suffix', 'band'])
+    datasets = [
+        ISCE2_DATASET('filt_topophase.unw.geo', 'unw_phase', 2),
+        ISCE2_DATASET('phsig.cor.geo', 'corr', 1),
+        ISCE2_DATASET('z.rdr.full.geo', 'dem', 1),
+        ISCE2_DATASET('filt_topophase.unw.conncomp.geo', 'conncomp', 1),
+        ISCE2_DATASET('filt_topophase.flat.geo', 'wrapped_phase', 1),
+    ]
+
+    for dataset in datasets:
```
jtherrmann, line 137: I think it might read a bit cleaner if we handled the `wrapped_phase` dataset outside of this loop (either before or after the loop), and removed it from the `datasets` list, so that this loop *only* performs the `gdal.Translate` calls.

src/hyp3_isce2/insar_tops_burst.py
```diff
@@ -126,6 +118,64 @@ def make_parameter_file(
...
+                srcDS=in_file,
+                bandList=[dataset.band],
+                format='GTiff',
+                noData=0,
+                creationOptions=['TILED=YES', 'COMPRESS=LZW', 'NUM_THREADS=ALL_CPUS'],
+            )
+
+    ds = gdal.Open(str(product_dir / 'filt_topophase.unw.geo'))
+    geotransform = ds.GetGeoTransform()
+    ds = None
```
jtherrmann, line 161: Would this accomplish the same thing? If so, then I think it would make the intention more explicit.
```suggestion
    del ds
```

jtherrmann: Or just:

```python3
geotransform = gdal.Open(str(product_dir / 'filt_topophase.unw.geo')).GetGeoTransform()
```

src/hyp3_isce2/insar_tops_burst.py
```diff
@@ -126,6 +118,64 @@ def make_parameter_file(
...
+
+    epsg = utm_from_lon_lat(geotransform[0], geotransform[3])
+    files = [str(x) for x in Path(product_name).glob('*.tif')]
```
jtherrmann, line 164:
```suggestion
    files = [str(path) for path in Path(product_name).glob('*.tif')]
```

jtherrmann: Alternatively, if this is just the result of all the `Translate` commands run above, then we should be able to construct the list of expected `tif` files without the need for a `glob`, correct?

src/hyp3_isce2/utils.py
```diff
@@ -0,0 +1,49 @@
+from osgeo import gdal
+
+
+def utm_from_lon_lat(lon: float, lat: float) -> int:
+    """Get the UTM zone epsg from a longitude and latitude.
```
jtherrmann, line 5:
```suggestion
    """Get the UTM zone EPSG code from a longitude and latitude.
```

src/hyp3_isce2/utils.py
```diff
@@ -0,0 +1,49 @@
...
+
+    Args:
+        lon: Longitude
+        lat: Latitude
+
+    Returns:
+        UTM zone epsg
```
jtherrmann, line 12:
```suggestion
        UTM zone EPSG code
```

src/hyp3_isce2/utils.py
```diff
@@ -0,0 +1,49 @@
...
+    """
+    hemisphere = 32600 if lat >= 0 else 32700
+    zone = int(lon // 6 + 30) % 60 + 1
+    return hemisphere + zone
```
jtherrmann, line 16: Where does this formula come from? Is it common knowledge, otherwise maybe worth linking to the source?

src/hyp3_isce2/insar_tops_burst.py
forrestfwilliams: Ooh I don't think I knew about `check_call`. Nice.

src/hyp3_isce2/insar_tops_burst.py
forrestfwilliams: Yes, that's correct.

src/hyp3_isce2/insar_tops_burst.py
forrestfwilliams: Yeah, I went back and forth on this. With your vote, I'll refactor to remove the `if` statement

src/hyp3_isce2/utils.py
forrestfwilliams: This comes from `hyp3-gamma`, but sure where they're from originally.

src/hyp3_isce2/insar_tops_burst.py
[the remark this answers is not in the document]
forrestfwilliams: unfortunately, no.

PR #50, opened 2023-05-18
Add `++omp-num-threads` argument

jtherrmann:
I tested this option with 4 threads and confirmed that I saw 4 threads being used in the log output, whereas I saw 8 threads without this option.

### DIFF
--- a/src/hyp3_isce2/__main__.py
+++ b/src/hyp3_isce2/__main__.py
@@ -2,6 +2,7 @@
 ISCE2 processing for HyP3
 """
 import argparse
+import os
 import sys
 from importlib.metadata import entry_points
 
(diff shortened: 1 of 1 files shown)


src/hyp3_isce2/__main__.py
```diff
@@ -21,12 +22,17 @@ def main():
         default='insar_tops_burst',
         help='Select the HyP3 entrypoint to use',  # HyP3 entrypoints are specified in `pyproject.toml`
     )
+    parser.add_argument('++omp-num-threads', type=int, help='The number of OpenMP threads to use for parallel regions')
```
forrestfwilliams, line 25: What happens if the user doesn't specify this? Do we need a default? If so, I'd suggest 8.

forrestfwilliams (approved):
LGTM. Just one comment.

src/hyp3_isce2/__main__.py
jtherrmann: No idea, I just copied the `hyp3-gamma` implementation. I'll merge for now and we can circle back to it before merging to `develop`, if we feel it's important.

jhkennedy: Not setting it lets ISCE2/OpenMP use *their* defaults, which I believe is the behavior we'd want.

PR #51, opened 2023-05-18
Create thumbnails and upload individual files

jtherrmann:
* Create thumbnail for browse image
* Upload individual files to S3
* Rename some variables to make the code more similar to `hyp3-gamma`

### DIFF
--- a/src/hyp3_isce2/insar_tops_burst.py
+++ b/src/hyp3_isce2/insar_tops_burst.py
@@ -13,6 +13,7 @@ from shutil import copyfile, make_archive
 
 from hyp3lib.aws import upload_file_to_s3
 from hyp3lib.get_orb import downloadSentinelOrbitFile
+from hyp3lib.image import create_thumbnail
 from osgeo import gdal
 
 from hyp3_isce2 import topsapp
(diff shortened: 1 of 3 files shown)


asjohnston-asf:
Generally woks how I expected. I see a 2048 pixel wide browse image, and a 100 pixel wide thumbnail image. All these files were uploaded to S3 with the correct `file_type` S3 Tags:
```
_conncomp.tif
_corr.tif
_dem.tif
.json
_unw_phase.png
_unw_phase.png.aux.xml
_unw_phase_thumb.png
_unw_phase.tif
_unw_phase.tif.aux.xml
_wrapped_phase.tif
```

Possibly out of scope for this PR, but I think we should remove the two sidecar `.aux.xml` files. I believe `_unw_phase.tif.aux.xml` contains the stats we generate to produce the browse image, and `_unw_phase.png.aux.xml` has geolocation information for the browse image. We don't include either for INSAR_GAMMA products.

asjohnston-asf (approved):

asjohnston-asf:
opened https://github.com/ASFHyP3/hyp3-isce2/pull/52 to address the `.aux.xml` files, but feel free to merge this and point that one at `develop` if you prefer

asjohnston-asf (approved):

Issue #53, opened 2023-05-19
Clean up logging


jtherrmann:
From @asjohnston-asf:

> not getting log output, only isce output, I think because https://github.com/ASFHyP3/hyp3-isce2/blob/develop/src/hyp3_isce2/insar_tops_burst.py#L30 sets `log` to the module logger, and https://github.com/ASFHyP3/hyp3-isce2/blob/develop/src/hyp3_isce2/insar_tops_burst.py#L193 only configures the root logger
>
> we have a mix of `log = logging.getLogger()` and `log = logging.getLogger(__name__)` in hyp3-gamma, I remember it being a mess to clean up there when I refactored RTC and I don't recall if we ever cleaned up InSAR

jtherrmann:
Actually, configuring the root logger also configures all the child loggers. What we have *should* work, except it seems that the root logger is already configured. From https://docs.python.org/3/library/logging.html#logging.basicConfig:

>This function does nothing if the root logger already has handlers configured, unless the keyword argument _force_ is set to `True`.

I confirmed that setting `force=True` makes logging work. But if the root logger has already been configured, maybe we don't want to mess with it. So still deciding on the best solution.

jtherrmann:
Hmm, that fixed the level issue, but it seems the root logger still has a pre-existing configuration so that `basicConfig` only works with `force=True`.

jtherrmann:
Alright, I got way too curious about where the root logger was being configured, but I tracked it down to here: https://github.com/isce-framework/isce2/blob/a5630c12030c6ffc7e5f4106471a3e7a9961b438/__init__.py#L37-L39

The formatter is specified here: https://github.com/isce-framework/isce2/blob/a5630c12030c6ffc7e5f4106471a3e7a9961b438/defaults/logging/logging.conf#L32

For future reference, I tracked it down by configuring the root logger at the beginning of our code:

```python3
logging.basicConfig(stream=sys.stdout, format='%(asctime)s - %(levelname)s - %(message)s', level=logging.INFO)
```

And then strategically placing `logging.info` calls until I found the exact `import` statement that was causing the log record format to change:

```python3
from isce.applications.topsApp import TopsInSAR
```

Anyway, now that my curiosity is satisfied, I think that either forcing our own logging configuration or just using the `isce2` configuration would both be reasonable options.

jtherrmann:
[...]

PR #56, opened 2023-05-19
Clean up logging

jtherrmann:
Fixes https://github.com/ASFHyP3/hyp3-isce2/issues/53

TODO:

- [ ] Confirm the following line appears in output: `ISCE2 TopsApp run completed successfully`


jtherrmann:
This PR uses the root logger everywhere, but after learning more about logging I've decided we do want to use child loggers. So closing in favor of a follow-up PR (incoming).

PR #59, opened 2023-05-19
Clean up logging and use GDAL exceptions everywhere

jtherrmann:
Fixes https://github.com/ASFHyP3/hyp3-isce2/issues/53

TODO:

- [x] Check if we're getting unwanted log output from `rasterio` and fix that if necessary (that's probably why we set the level in `dem.py` originally).
- [x] Confirm our `log.info` statements are working as expected.

### DIFF
--- a/src/hyp3_isce2/logging.py
+++ b/src/hyp3_isce2/logging.py
@@ -1,0 +1,13 @@
+import logging
+import sys
+
+
+def configure_root_logger() -> None:
+    # isce2 sets the root logger level to DEBUG, so we re-configure it here.
+    # See: https://github.com/ASFHyP3/hyp3-isce2/issues/53
+    logging.basicConfig(
+        stream=sys.stdout,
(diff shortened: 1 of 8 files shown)


src/hyp3_isce2/logging.py
```diff
@@ -0,0 +1,13 @@
+import logging
+import sys
+
+
+def configure_root_logger() -> None:
+    # isce2 sets the root logger level to DEBUG, so we re-configure it here.
+    # See: https://github.com/ASFHyP3/hyp3-isce2/issues/53
```
forrestfwilliams, line 7: Maing this comment as a docstring makes it easier for people to access it.
```suggestion
def configure_root_logger() -> None:
    """Re-configures the root level logger since ISCE2 sets it to DEBUG.
    
    See: https://github.com/ASFHyP3/hyp3-isce2/issues/53 for context.
    """
```

forrestfwilliams (commented):
LGTM, just one suggestion.

forrestfwilliams (approved):

PR #71, opened 2023-05-26
Fix Parameter File

AndrewPlayer3:
The parameter file expected the full SLC name instead of the burst name. 

TODO:
- [x] Find the .SAFE directories with only the burst name provided.

Also adds new example command to the README.


jtherrmann (approved):
Looks great. It would be nice to have test coverage for this functionality, but I think we will end up needing to write tests for the entirety of `make_parameter_file` at some point, and perhaps breaking the function up to make testing easier. So no need to add tests as part of this PR.

PR #72, opened 2023-05-26
Burst not found error message

AndrewPlayer3:
Added a helpful error message when CMR cannot find one or both of the requested bursts.

### DIFF
--- a/src/hyp3_isce2/insar_tops_burst.py
+++ b/src/hyp3_isce2/insar_tops_burst.py
@@ -68,6 +68,15 @@ def insar_tops_burst(
     )
     results = asf_search.search(opts=opts)
 
+    ref_not_found = reference_scene not in results.__str__()
+    sec_not_found = secondary_scene not in results.__str__()
+    if ref_not_found and sec_not_found:
+        raise RuntimeError(f'ASF Search failed to find both {reference_scene} and {secondary_scene}.')
+    elif ref_not_found:
+        raise RuntimeError(f'ASF Search failed to find {reference_scene}.')
+    elif sec_not_found:
+        raise RuntimeError(f'ASF Search failed to find {secondary_scene}.')
+
     ref_params = BurstParams(
         results[0].umm['InputGranules'][0].split('-')[0],
         results[0].properties['burst']['subswath'],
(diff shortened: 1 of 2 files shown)


src/hyp3_isce2/insar_tops_burst.py
```diff
@@ -68,6 +68,15 @@ def insar_tops_burst(
     )
     results = asf_search.search(opts=opts)
 
+    ref_not_found = reference_scene not in results.__str__()
+    sec_not_found = secondary_scene not in results.__str__()
```
jtherrmann, line 72: I don't recall the exact interface for search results but I would guess there's a more idiomatic way to check if the scenes are in the results. If not, then I would at least suggest changing this to `str(results)`.

src/hyp3_isce2/insar_tops_burst.py
```diff
@@ -68,6 +68,15 @@ def insar_tops_burst(
...
+    if ref_not_found and sec_not_found:
+        raise RuntimeError(f'ASF Search failed to find both {reference_scene} and {secondary_scene}.')
+    elif ref_not_found:
+        raise RuntimeError(f'ASF Search failed to find {reference_scene}.')
+    elif sec_not_found:
+        raise RuntimeError(f'ASF Search failed to find {secondary_scene}.')
```
jtherrmann, line 78: We should use a more specific exception type, but not sure what that might be off the top of my head.

src/hyp3_isce2/insar_tops_burst.py
```diff
@@ -68,14 +70,14 @@ def insar_tops_burst(
...
+    sec_not_found = secondary_scene not in str(results)
     if ref_not_found and sec_not_found:
-        raise RuntimeError(f'ASF Search failed to find both {reference_scene} and {secondary_scene}.')
+        raise ASFSearchError(f'ASF Search failed to find both {reference_scene} and {secondary_scene}.')
     elif ref_not_found:
-        raise RuntimeError(f'ASF Search failed to find {reference_scene}.')
+        raise ASFSearchError(f'ASF Search failed to find {reference_scene}.')
     elif sec_not_found:
-        raise RuntimeError(f'ASF Search failed to find {secondary_scene}.')
+        raise ASFSearchError(f'ASF Search failed to find {secondary_scene}.')
```
jtherrmann: I don't think we should explicitly raise `ASFSearchError`, as I believe that's an error that only the `asf_search` module itself should raise. If we can't think of a more appropriate exception, [`ValueError`](https://docs.python.org/3/library/exceptions.html#ValueError) is probably a safe bet.

PR #76, opened 2023-05-31
Update readme

cirrusasf:


src/hyp3_isce2/metadata/templates/insar_burst/readme.md.txt.j2
```diff
@@ -101,36 +115,31 @@ https://spacedata.copernicus.eu/collections/copernicus-digital-elevation-model.
...
+## 2. Unwrapped Interferogram
 
-The unwrapped interferogram uses a reference point to convert the wrapped 2-pi scale into a continuous scale (of multiples of pi), and is the most important and complicated step in InSAR processing. The Minimum Cost Flow (MCF) approach (https://www.gamma-rs.ch/uploads/media/2002-5_TR_Phase_Unwrapping.pdf) was used for phase unwrapping this product, and the reference point was set to the location of the pixel with the highest 9-pixel neighborhood coherence value. The reference point for this interferogram is located at {{ ref_point_coords['lat'] }} degrees latitude and {{ ref_point_coords['lon'] }} degrees longitude in WGS84 geographic coordinates.
+The unwrapped interferogram uses a reference point to convert the wrapped 2-pi scale into a continuous scale (of multiples of pi), and is the most important and complicated step in InSAR processing. The Minimum Cost Flow (MCF) approach (https://www.gamma-rs.ch/uploads/media/2002-5_TR_Phase_Unwrapping.pdf) was used for phase unwrapping this product, and the reference point was set to the location of the pixel with the highest 9-pixel neighborhood coherence value.
```
hjkristenson, line 142: - Are we not indicating the reference point location for these products?
- I thought that ISCE doesn't use MCF 
- Are we using the same approach for setting the reference point that we use for the GAMMA products?

src/hyp3_isce2/insar_tops_burst.py
```diff
@@ -363,6 +367,37 @@ def main():
...
+    payload['reference_granule_name'] = args.reference_scene
+    payload['secondary_granule_name'] = args.secondary_scene
+    payload['swath_number'] = args.swath_number
+    payload['polarization'] = args.polarization
+    payload['reference_burst_number'] = args.reference_burst_number
+    payload['secondary_burst_number'] = args.secondary_burst_number
+    payload['processing_date'] = datetime.now(timezone.utc)
+    payload['range_looks'] = args.range_looks
+    payload['azimuth_looks'] = args.azimuth_looks
+    payload['dem_name'] = 'GLO-30'
```
hjkristenson, line 382: Because we just use the one DEM, that changes the way we deal with DEM-related information. Is there a place that we define the pixel spacing of the GLO-30 DEM, or should we just hard-code that value on line 199 of the metadata template?

hjkristenson (approved):
I've gone through the template content and added a few comments. They can be addressed after merging, but I wanted to note them while I was thinking about it.

src/hyp3_isce2/insar_tops_burst.py
cirrusasf: previous I use marco to get the DEM pixel spacing of GLO-30 DEM, since we do not want to use the macro, we need hard code, or we can leave it as jinjn replacement, and use the payload to replace it. Either way is OK.

src/hyp3_isce2/metadata/templates/insar_burst/readme.md.txt.j2
cirrusasf: We do not indicate the reference point, and I don't know how the ISCE2 determine the reference point. May check the ISCE2 document to find out.

src/hyp3_isce2/insar_tops_burst.py
hjkristenson: If it's okay to just add a payload entry, that's probably the most transparent. It probably doesn't make much difference, as if we ever change the DEM, we'll have to completely overhaul the DEM description sections, but it somehow seems more robust to me.

hjkristenson: Resolved with https://github.com/ASFHyP3/hyp3-isce2/pull/76/commits/d397179508442e956a2de8ee4ce20a64c3f85e6f

hjkristenson (approved):

PR #79, opened 2023-06-01
implement burst.get_burst_params() to resolve reference/secondary ordering bug

asjohnston-asf:
TODO:
- [x] mock search requests in `test_get_burst_params()`

### DIFF
--- a/src/hyp3_isce2/burst.py
+++ b/src/hyp3_isce2/burst.py
@@ -354,8 +354,8 @@ def get_burst_params(scene_name: str) -> BurstParams:
         raise ValueError(f'ASF Search found multiple results for {scene_name}.')
 
     return BurstParams(
-        results[0].umm['InputGranules'][0].split('-')[0],
-        results[0].properties['burst']['subswath'],
-        results[0].properties['polarization'],
-        results[0].properties['burst']['burstIndex'],
+        granule=results[0].umm['InputGranules'][0].split('-')[0],
+        swath=results[0].properties['burst']['subswath'],
+        polarization=results[0].properties['polarization'],
+        burst_number=results[0].properties['burst']['burstIndex'],
     )
--- a/src/hyp3_isce2/burst.py
+++ b/src/hyp3_isce2/burst.py
@@ -8,6 +8,7 @@ from dataclasses import dataclass
 from pathlib import Path
 from typing import Iterator, List, Optional, Tuple, Union
 
+import asf_search
 import isce  # noqa: F401
 import requests
 from isceobj.Sensor.TOPS.Sentinel1 import Sentinel1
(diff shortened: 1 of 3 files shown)


src/hyp3_isce2/burst.py
```diff
@@ -337,3 +338,24 @@ def get_product_name(
...
+    return asf_search.search(product_list=[scene_name], opts=opts)
+
+
+def get_burst_params(scene_name: str) -> BurstParams:
+    results = search_cmr_uat(scene_name)
+
+    if len(results) == 0:
+        raise ValueError(f'ASF Search failed to find {scene_name}.')
+    if len(results) > 1:
+        raise ValueError(f'ASF Search found multiple results for {scene_name}.')
```
AndrewPlayer3, line 354: Not blocking, but personally I think this may be better as a single if statement that includes the number of results. Like:
```suggestion
    if len(results) != 1:
        raise ValueError(f'ASF Search found {len(results)} results for {scene_name}.')
```
Then both the tests would have regexes matching `found 0 results` and `found 2 results`.

src/hyp3_isce2/burst.py
```diff
@@ -337,3 +338,24 @@ def get_product_name(
         The name of the interferogram product.
     """
     return f'{reference_scene}x{secondary_scene}'
+
+
+def search_cmr_uat(scene_name: str) -> asf_search.ASFSearchResults:
```
forrestfwilliams, line 343: This can be a TODO for the future, but in the period of time when CMR is NOT back-populated it would be great to check the date of the burst up-front and raise an error if it's date is before burst ingesting began. Something like:

`ValueError('Bursts before [date] are not currently available from ASF')`

tests/test_burst.py
```diff
@@ -88,3 +90,69 @@ def test_get_region_of_interest(tmp_path, orbit):
 
 def test_get_product_name():
     assert burst.get_product_name('A', 'B') == 'AxB'
+
+
+def mock_asf_search_results(
```
forrestfwilliams, line 95: nice use of mocking

forrestfwilliams (approved):
LGTM! You can add TODO for my comment above, but it's not a big deal either way.

src/hyp3_isce2/burst.py
jtherrmann: 100% agree. But not going to act on it now because I don't want to change the tests.

src/hyp3_isce2/burst.py
jtherrmann: Stuck this in an issue: https://github.com/ASFHyP3/hyp3-isce2/issues/84

AndrewPlayer3 (approved):

PR #83, opened 2023-06-02
Switch to use dem.crop as dem source

forrestfwilliams:
Geocoding `z.rdr.full` led to errors in the output DEM. This PR switches to using `dem.crop` by add a geotransform to it.


AndrewPlayer3 (approved):
Looks great!

PR #92, opened 2023-06-07
Clean up burst readme generation, fix ref/sec scene order

jtherrmann:
* Move all of the code for generating the burst README into a separate helper function
* Delete unused functions from the `metadata.util` module
* Fix a bug in which the burst README does not preserve reference/secondary granule name order
   * As seen in this test job: https://hyp3-enterprise-test.asf.alaska.edu/jobs/9140150a-861e-4513-bb3a-ac04cc5860f1

TODO:

- [x] Run the plugin locally to confirm the README generation still works.

### DIFF
--- a/src/hyp3_isce2/burst.py
+++ b/src/hyp3_isce2/burst.py
@@ -337,6 +337,8 @@ def get_product_name(
     Returns:
         The name of the interferogram product.
     """
+    # If this changes, we will also need to update the burst product README template,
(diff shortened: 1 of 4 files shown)


cirrusasf (approved):
check the change made. everything looks good to me.

forrestfwilliams (approved):
LGTM. Runs in my local environment, and produces the readme as expected.

PR #101, opened 2023-06-29
Control resolution for product

cirrusasf:


src/hyp3_isce2/dem.py
```diff
@@ -58,16 +59,39 @@ def buffer_extent(extent: list, buffer: float) -> list:
     ]
 
 
+def utm_from_lon_lat(lon: float, lat: float) -> int:
```
AndrewPlayer3, line 62: Not blocking, but it might be good for utm_from_lon_lat and get_dem_resolution to have comments.

src/hyp3_isce2/dem.py
```diff
@@ -58,16 +59,39 @@ def buffer_extent(extent: list, buffer: float) -> list:
...
+    latc = extent[3]
+
+    epsg_code = utm_from_lon_lat(lonc, latc)
+    myprj1 = pyproj.Transformer.from_crs(4326, epsg_code, always_xy=True)
+    myprj2 = pyproj.Transformer.from_crs(epsg_code, 4326, always_xy=True)
+    xc, yc = myprj1.transform(lonc, latc)
+    x2 = xc + res
+    y2 = yc - res
+    lon2, lat2 = myprj2.transform(x2, y2)
+    return abs(lon2 - lonc), abs(lat2 - latc)
```
AndrewPlayer3, line 80: Since we aren't using the latitude resolution anymore, it might be good to remove it from the return statement.

cirrusasf: good idea.

AndrewPlayer3 (approved):
Looks good now!

asjohnston-asf:
I confirmed resampling the dem prior to generating the interferogram has the desired impact on runtime on my laptop:
- 20x4: 3m28.131s
- 10x2: 4m13.908s
- 5x1: 8m11.577s

asjohnston-asf:
There's a subtle difference between the terms "pixel size" and "resolution". We should be specific and use `pixel_size` where ever we can for the new parameters/variables introduced in this PR.

"resolution" is roughly "how far apart do two objects need to be to distinguish them in the image", which is often related to pixel size, but not necessarily the same. You could resample an image to have much smaller pixels, but that doesn't improve the resolution of the image, for example.

src/hyp3_isce2/insar_tops_burst.py
```diff
@@ -347,11 +349,19 @@ def translate_outputs(isce_output_dir: Path, product_name: str):
             file,
             dstSRS=f'epsg:{epsg}',
             creationOptions=['TILED=YES', 'COMPRESS=LZW', 'NUM_THREADS=ALL_CPUS'],
+            xRes=res,
+            yRes=res,
+            targetAlignedPixels=True
```
asjohnston-asf, line 354: @forrestfwilliams @hjkristenson any opinions on which [resampling method](https://gdal.org/programs/gdalwarp.html#cmdoption-gdalwarp-r) to use when reprojecting these geotiffs to their final UTM projection and pixel size?

asjohnston-asf:
alternate implementation released (see hyp3-isce2 v0.6.0)

PR #108, opened 2023-07-07
use native dem (30m) for the calculation, resample the product with p…

cirrusasf:
…ixel_size determined by looks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/)
 and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.5.1]
+### Changed
+* Pixel size of output products is set to be the same resolution as Geocoding DEM
 
 ## [0.5.0]
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [0.6.0]
 ### Changed
-* Pixel size of output products is set to be the same resolution as Geocoding DEM
+* Pixel sizes of output GeoTIFFs now better reflect their resolution: 80m for 20x4 looks, 40m for 10x2 looks, and 20m for 5x1 looks.
 
 ## [0.5.0]
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/)
 and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [0.5.1]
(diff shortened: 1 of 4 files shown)


src/hyp3_isce2/insar_tops_burst.py
```diff
@@ -242,6 +244,7 @@ def make_parameter_file(
         'Azimuth bandpass filter: no\n',
         f'DEM source: {dem_name}\n',
         f'DEM resolution (m): {dem_resolution}\n',
+        f'Pixel Size (m): {pixel_size}\n',
```
asjohnston-asf, line 247: I vote against adding the pixel size to the parameter .txt file unless someone has a compelling use case for it. Pixel size is a self-describing property of the output geotiffs, so I don't think the extra code to include it in the txt file is justified. Thoughts?

src/hyp3_isce2/insar_tops_burst.py
```diff
@@ -371,6 +382,8 @@ def main():
 
     args = parser.parse_args()
 
+    pixel_size = get_pixel_size(args.looks)
```
asjohnston-asf, line 385: Good practice is to wait to define variables until they're needed to reduce the number of lines a reader needs to keep in their head at once. This is better moved to just before the `translate_outputs()` call on line 414.

src/hyp3_isce2/insar_tops_burst.py
cirrusasf: will delete the pixel size in the txt file

src/hyp3_isce2/insar_tops_burst.py
```diff
@@ -414,7 +424,8 @@ def main():
         secondary_scene=secondary_scene,
         azimuth_looks=azimuth_looks,
         range_looks=range_looks,
-        swath_number=swath_number
+        swath_number=swath_number,
+        pixel_size=pixel_size
```
asjohnston-asf, line 428:
```suggestion
        swath_number=swath_number
```

src/hyp3_isce2/insar_tops_burst.py
```diff
@@ -167,6 +167,7 @@ def make_parameter_file(
         range_looks: Number of range looks
         dem_name: Name of the DEM that is use
         dem_resolution: Resolution of the DEM
+        pixel_size: pixel spacing of output product
```
asjohnston-asf, line 170:
```suggestion
```

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/)
 and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.6.0]
+### Changed
+* Pixel size of output products is set to be the same resolution as Geocoding DEM
```
asjohnston-asf, line 11:
```suggestion
* Pixel sizes of output GeoTIFFs now better reflect their resolution: 80m for 20x4 looks, 40m for 10x2 looks, and 20m for 5x1 looks.
```

asjohnston-asf:
With the changelog tweak I'll be content to move this into develop and release shortly after. I'll want to document that we're likely losing some resolution for 5x1 products by geocoding at ~27m before resampling to 20m, but I'm content to revisit that issue later.

asjohnston-asf (approved):

PR #137, opened 2023-09-08
Clean up burst workflow before going to production

jtherrmann:
I've gone through `insar_tops_burst.py` thoroughly. I don't think there's anything more to improve there.

I did a quick pass through `burst.py` and `topsapp.py` to look for obvious things, but haven't done a thorough read-through.

I haven't looked at any of the other source files yet.

TODO:

- [x] Review `insar_tops_burst.py`, and the source files that it imports, for functions without unit tests (unless they would require too much mocking)
- [x] Add missing unit tests
- [x] Finish going through `burst.py`, `topsapp.py`, and the other source files to look for general code quality, PyCharm linter errors, and add any missing type hints.
- [x] Search for remaining `TODO` and `FIXME` items


src/hyp3_isce2/burst.py
AndrewPlayer3, line 1: I feel like reference_name in the arguments should be reference_scene, since it's called reference_scene everywhere else. Otherwise, we should rename to reference_name everywhere.

AndrewPlayer3 (approved):

PR #138, opened 2023-09-08
Unit test for make_browse_image

AndrewPlayer3:


jtherrmann:
I reviewed the test browse image to verify it is 2048x2048, 8-bit grayscale, and looks similar to `tests/data/test_geotiff.tif` (displayed in QGIS).

PR #151, opened 2023-10-30
Fix call to `imageMath.py`

jtherrmann:
Currently, if I build the conda environment locally and run the example command from the README:

```
python -m hyp3_isce2 ++process insar_tops_burst \
  S1_136231_IW2_20200604T022312_VV_7C85-BURST \
  S1_136231_IW2_20200616T022313_VV_5D11-BURST \
  --looks 20x4 \
  --apply-water-mask True
```

It fails with: `sh: 1: ImageMath.py: not found`.  Using `subprocess` instead of `os.system` should resolve the path issue.

Edit: Nope, still fails with the same problem: `FileNotFoundError: [Errno 2] No such file or directory: 'ImageMath.py'`

Edit 2: It was a typo, should be `imageMath.py`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/)
 and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.8.1]
+### Fixed
+* Fixed a typo in the call to `imageMath.py`.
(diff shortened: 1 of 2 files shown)


forrestfwilliams:
Have you checked that `$ISCE_HOME/applications` is in your `PATH`? [These lines](https://github.com/ASFHyP3/hyp3-isce2/blob/a3f43335e459c27b0ab4d684332c92d569caa5f5/src/hyp3_isce2/insar_tops_burst.py#L54C12-L54C12) make sure this is the case, but maybe there's an issue there, or maybe you're trying to use image_math.py before these lines are run.

jhkennedy (approved):

PR #156, opened 2023-11-10
Ruff proposal

jhkennedy:
This updates ruff config in-line with the updates to the ruff action proposed in ASFHyP3/actions#95 and applies the lining auto fixes. 

In particular, adopting the additional checks would result in needing to make quite a few changes to satisfy the linter:
```
$ ruff check --statistics .
56      ANN001  [ ] Missing type annotation for function argument `array`
56      ANN201  [ ] Missing return type annotation for public function `check_correctness_of_resample`
55      D103    [ ] Missing docstring in public function
32      D415    [*] First line should end with a period, question mark, or exclamation point
20      D100    [ ] Missing docstring in public module
13      PTH123  [ ] `open()` should be replaced by `Path.open()`
10      ANN101  [ ] Missing type annotation for `self` in method
 8      D205    [ ] 1 blank line required between summary line and description
 6      ANN204  [ ] Missing return type annotation for special method `__enter__`
 3      PTH107  [ ] `os.remove()` should be replaced by `Path.unlink()`
 3      D107    [ ] Missing docstring in `__init__`
 2      PTH113  [ ] `os.path.isfile()` should be replaced by `Path.is_file()`
 2      D105    [ ] Missing docstring in magic method
 2      D200    [*] One-line docstring should fit on one line
 2      D417    [ ] Missing argument description in the docstring for `create_water_mask`: `input_image`
 1      ANN003  [ ] Missing type annotation for `**options`
 1      ANN202  [ ] Missing return type annotation for private function `_download_platform`
 1      PTH109  [ ] `os.getcwd()` should be replaced by `Path.cwd()`
 1      PTH120  [ ] `os.path.dirname()` should be replaced by `Path.parent`
 1      PTH207  [ ] Replace `glob` with `Path.glob` or `Path.rglob`
 1      D104    [ ] Missing docstring in public package
```

### DIFF
--- a/src/hyp3_isce2/burst.py
+++ b/src/hyp3_isce2/burst.py
@@ -177,7 +177,7 @@ def download_burst(asf_session: requests.Session, burst_params: BurstParams, out
     return Path(out_file)
 
 
-def spoof_safe(burst: BurstMetadata, burst_tiff_path: Path, base_path: Path = Path('.')) -> Path:
+def spoof_safe(burst: BurstMetadata, burst_tiff_path: Path, base_path: Path = Path.cwd()) -> Path:
     """Spoof a Sentinel-1 SAFE file for a burst.
 
     The created SAFE file will be saved to the base_path directory. The SAFE will have the following structure:
@@ -223,12 +223,12 @@ def get_isce2_burst_bbox(params: BurstParams, base_dir: Optional[Path] = None) -
     """Get the bounding box of a Sentinel-1 burst using ISCE2.
     Using ISCE2 directly ensures that the bounding box is the same as the one used by ISCE2 for processing.
 
-    args:
+    Args:
         params: The burst parameters.
         base_dir: The directory containing the SAFE file.
             If base_dir is not set, it will default to the current working directory.
 
-    returns:
+    Returns:
         The bounding box of the burst as a shapely.geometry.Polygon object.
     """
     if base_dir is None:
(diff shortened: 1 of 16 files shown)


src/hyp3_isce2/burst.py
```diff
@@ -177,7 +177,7 @@ def download_burst(asf_session: requests.Session, burst_params: BurstParams, out
     return Path(out_file)
 
 
-def spoof_safe(burst: BurstMetadata, burst_tiff_path: Path, base_path: Path = Path('.')) -> Path:
+def spoof_safe(burst: BurstMetadata, burst_tiff_path: Path, base_path: Path = Path.cwd()) -> Path:
```
jhkennedy, line 180: Ruff does not like `Path('.')` and would prefer you use either `Path()` or `Path.cwd()`. There were some gotcha's here around when `'.'` was resolved to an actual path, so we'll want to make sure that `Path.cwd()` doesn't hit those again....

src/hyp3_isce2/stripmapapp_alos.py
```diff
@@ -123,7 +123,7 @@ def run_stripmapapp(dostep: str = '', start: str = '', end: str = '', config_xml
         ValueError: If the step is not a valid step (see TOPSAPP_STEPS)
     """
     if not config_xml.exists():
-        raise IOError(f'The config file {config_xml} does not exist!')
+        raise OSError(f'The config file {config_xml} does not exist!')
```
jhkennedy, line 126: `IOError` is just a legacy/backwards compatible alias for `OSError` so `OSError` is preferred.

PR #157, opened 2023-11-10
Changed to hyp3-gamma's New Water Masking Method

AndrewPlayer3:
This PR changes the water mask generation code to use hyp3-gamma's new method, rather than the current parque method. This is due to issues that arise around the boundaries of the parque partitions.


forrestfwilliams:
@AndrewPlayer3 I think we try to find a way to resolve this issue that still allows us to use the parquet file. When running the shapefile version of the burst workflow on a local machine, creating the watermask is very slow and can take up to 30-40% of the total processing time.

While we typically run these workflow in AWS where downloads are faster, I think we need to consider the local use case here because this is the only workflow in the community for creating single-burst InSAR and people will likely want to use it with their own modifications.

Issue #176, opened 2024-01-11
botocore.exceptions.ConnectTimeoutError and lots of DEBUG outputs


scottyhq:
### The bug
Running hyp3-isce2 (0.9.2) locally I'm seeing a lot of initial logging messages related to botocore looking for cloud credentials. It looks like there is a traceback related to looking for instance metadata (would only work on AWS), but in any case the code ultimately succeeds successfully.

### To Reproduce
```bash
docker run -it --rm \
    -e EARTHDATA_USERNAME=$EARTHDATA_USERNAME \
    -e EARTHDATA_PASSWORD=$EARTHDATA_PASSWORD \
    -e ESA_USERNAME=$ESA_USERNAME \
    -e ESA_PASSWORD=$ESA_PASSWORD \
    ghcr.io/asfhyp3/hyp3-isce2:0.9.2 \
        ++process insar_tops_burst \
        S1_136231_IW2_20200604T022312_VV_7C85-BURST \
        S1_136231_IW2_20200616T022313_VV_5D11-BURST \
        --looks 20x4 \
        --apply-water-mask False
```

 ### Additional context

```pytb
2024-01-11 03:07:22,367 - botocore.credentials - DEBUG - Looking for credentials via: iam-role
2024-01-11 03:07:22,368 - urllib3.connectionpool - DEBUG - Starting new HTTP connection (1): 169.254.169.254:80
2024-01-11 03:07:23,370 - botocore.utils - DEBUG - Caught retryable HTTP exception while making metadata service request to http://169.254.169.254/latest/api/token: Connect timeout on endpoint URL: "http://169.254.169.254/latest/api/token"
Traceback (most recent call last):
  File "/opt/conda/envs/hyp3-isce2/lib/python3.11/site-packages/urllib3/connection.py", line 174, in _new_conn
    conn = connection.create_connection(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/envs/hyp3-isce2/lib/python3.11/site-packages/urllib3/util/connection.py", line 95, in create_connection
    raise err
  File "/opt/conda/envs/hyp3-isce2/lib/python3.11/site-packages/urllib3/util/connection.py", line 85, in create_connection
    sock.connect(sa)
TimeoutError: timed out
```

<details>
  <summary>full traceback</summary>

```pytb
[...]

PR #177, opened 2024-01-12
Workaround isce2 setting root logger to debug

scottyhq:
Fixes https://github.com/ASFHyP3/hyp3-isce2/issues/176

Ideally, the fix would be to isce2, but it's been a known problem for a while (https://github.com/isce-framework/isce2/issues/258)...

### DIFF
--- a/src/hyp3_isce2/insar_tops_burst.py
+++ b/src/hyp3_isce2/insar_tops_burst.py
@@ -12,9 +12,6 @@ from shutil import copyfile, make_archive
 from typing import Optional
 
 import isce
-# https://github.com/isce-framework/isce2/issues/258
-root_logger = logging.getLogger()
-root_logger.setLevel('WARNING')
 from hyp3lib.aws import upload_file_to_s3
 from hyp3lib.get_orb import downloadSentinelOrbitFile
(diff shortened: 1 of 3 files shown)


src/hyp3_isce2/insar_tops_burst.py
```diff
@@ -12,6 +12,9 @@
 from typing import Optional
 
 import isce
+# https://github.com/isce-framework/isce2/issues/258
+root_logger = logging.getLogger()
+root_logger.setLevel('WARNING')
```
jhkennedy, line 17: We added an import of `isce` to `hyp3_isce2.__init__` to ensure no matter where we are in this package, the ISCE2 stuff is set up correctly. Can you move this there?
https://github.com/ASFHyP3/hyp3-isce2/blob/develop/src/hyp3_isce2/__init__.py#L6-L13

jhkennedy (commented):
Nice! I agree this should be handled by ISCE2 but that issue is approaching 3 years old at this point, so I don't see that happening anytime soon.

jhkennedy (approved):
Looks great! I just added a changelog entry, fixed a listing issue, and added some more info to the logging comment.

PR #187, opened 2024-01-24
New Water Masks

AndrewPlayer3:
This PR adds and modifies code to support the updated raster water masks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/)
 and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.10.0]
+### Added
+* Support for a new water masking dataset based off of OpenStreetMaps and ESA WorldCover data.
 
 ## [0.9.3]
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ## [0.10.0]
 ### Added
 * Support for a new water masking dataset based off of OpenStreetMaps and ESA WorldCover data.
+### Removed
+* Polygon processing functions: `split_geometry_on_antimeridian` and `get_envelope_wgs84` from `water_mask.py`.
 
 ## [0.9.3]
 ### Changed
--- a/src/hyp3_isce2/metadata/templates/insar_burst/readme.md.txt.j2
+++ b/src/hyp3_isce2/metadata/templates/insar_burst/readme.md.txt.j2
@@ -280,13 +280,9 @@ process caused by invalid coherence over water bodies. The water mask will also 
 in the output products. This product {{ "has" if apply_water_mask else "has not" }} 
 had the water mask applied.
 
-The water mask is generated using the Global Self-consistent, Hierarchical, High-resolution Geography Database (GSHHG)
-dataset (https://www.ngdc.noaa.gov/mgg/shorelines). To generate the global mask, we combined the full-resolution
-L1 and L5 datasets, and removed the L2 dataset minus the L3 dataset. The L1 dataset is the boundary between land
-and ocean, excluding Antarctica, and the L5 dataset is the boundary between Antarctic ice and ocean. The L2
-dataset is the boundary between lakes and land, and the L3 dataset is the boundary between islands and the
-lakes they are within. The GSHHG dataset was last updated in 2017, so there may be discrepancies where
-shorelines have changed.
+The water mask is generated using data from OpenStreetMaps and/or ESA WorldCover depeding on location. Areas within 
+Canada, Alaska, and Russia are primarily covered by ESA WorldCover data, while the rest of the world is covered
+by OpenStreetMaps data. Water masks were previously generated from the GSHHG dataset.
 
 *************
 # Burst InSAR Processing #
--- a/src/hyp3_isce2/water_mask.py
+++ b/src/hyp3_isce2/water_mask.py
@@ -4,8 +4,6 @@ from os import system
 import numpy as np
 
 from osgeo import gdal
-from rasterio import open as ropen
-from rasterio.mask import mask
 
 gdal.UseExceptions()
 
@@ -14,7 +12,7 @@ TILE_PATH = '/vsicurl/https://asf-dem-west.s3.amazonaws.com/WATER_MASK/TILES/'
 
 def get_corners(filename):
     """Get all four corners of the given image: [upper_left, bottom_left, upper_right, bottom_right].
-    
+
     Args:
         filename: The path to the input image.
     """
@@ -33,7 +31,7 @@ def get_corners(filename):
 
 def coord_to_tile(coord: tuple(float, float)) -> str:
     """Get the filename of the tile which encloses the inputted coordinate.
-    
+
     Args:
         coord: The (lon, lat) tuple containing the desired coordinate.
     """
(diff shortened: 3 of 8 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/)
 and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.10.0]
+### Added
```
forrestfwilliams, line 10: I think we should also add information on the `water_mask.py` functionality that has been remove with the switch from a polygon to raster approach. Could you add some text to this effect?

src/hyp3_isce2/water_mask.py
```diff
@@ -1,55 +1,71 @@
...
 
+TILE_PATH = '/vsicurl/https://asf-dem-west.s3.amazonaws.com/WATER_MASK/TILES/'
 
-def split_geometry_on_antimeridian(geometry: dict):
-    geometry_as_bytes = json.dumps(geometry).encode()
-    cmd = ['ogr2ogr', '-wrapdateline', '-datelineoffset', '20', '-f', 'GeoJSON', '/vsistdout/', '/vsistdin/']
-    geojson_str = subprocess.run(cmd, input=geometry_as_bytes, stdout=subprocess.PIPE, check=True).stdout
-    return json.loads(geojson_str)['features'][0]['geometry']
 
+def get_corners(filename, tmp_path: str = ''):
```
forrestfwilliams, line 12: Having `tmp_path` be an empty strings seems like it could cause issues. Could we make it an `Optional[Path]` with a default of `None`?

src/hyp3_isce2/water_mask.py
```diff
@@ -1,55 +1,71 @@
...
+    """Get all four corners of the given image: [upper_left, bottom_left, upper_right, bottom_right].
 
-def get_envelope_wgs84(input_image: str):
-    """Get the envelope around a GeoTIFF.
     Args:
-        input_image: The path to the desired GeoTIFF, as a string.
-    Returns:
-        envelope_wgs84_gdf: The WGS84 envelope around the GeoTIFF, as a GeoDataFrame.
+        filename: The path to the input image.
```
forrestfwilliams, line 16: Could you add an entry for `tmp_path`?

src/hyp3_isce2/metadata/templates/insar_burst/readme.md.txt.j2
```diff
@@ -280,13 +280,9 @@ process caused by invalid coherence over water bodies. The water mask will also
...
 had the water mask applied.
 
-The water mask is generated using the Global Self-consistent, Hierarchical, High-resolution Geography Database (GSHHG)
-dataset (https://www.ngdc.noaa.gov/mgg/shorelines). To generate the global mask, we combined the full-resolution
-L1 and L5 datasets, and removed the L2 dataset minus the L3 dataset. The L1 dataset is the boundary between land
-and ocean, excluding Antarctica, and the L5 dataset is the boundary between Antarctic ice and ocean. The L2
-dataset is the boundary between lakes and land, and the L3 dataset is the boundary between islands and the
-lakes they are within. The GSHHG dataset was last updated in 2017, so there may be discrepancies where
-shorelines have changed.
+The water mask is generated using data from OpenStreetMaps and/or ESA WorldCover depeding on location. Areas within
```
forrestfwilliams, line 283: When your code eventually lives somewhere, we should link to it here.

tests/test_water_mask.py
```diff
@@ -1,127 +1,63 @@
...
+    )
+    case_2 = (
+        'tests/data/test_geotiff.tif',
+        ['/vsicurl/https://asf-dem-west.s3.amazonaws.com/WATER_MASK/TILES/n30w120.tif']
+    )
+    assert water_mask.get_tiles(case_1[0], tmp_path=str(tmp_path)) == case_1[1]
+    assert water_mask.get_tiles(case_2[0], tmp_path=str(tmp_path)) == case_2[1]
+
+
+def test_create_water_mask(tmp_path):
```
forrestfwilliams, line 56: I'd prefer not downloading production data during a test. I know we were doing this before your changes, but I think we have a chance to improve here. Could we try not doing this by breaking `create_water_mask` into some smaller functions and mocking the results of the first `Warp` call?

src/hyp3_isce2/water_mask.py
```diff
@@ -59,36 +75,50 @@ def create_water_mask(input_image: str, output_image: str, gdal_format='GTiff'):
...
+        cropToCutline=True,
+        xRes=pixel_size,
+        yRes=pixel_size,
+        targetAlignedPixels=True,
+        dstSRS='EPSG:4326',
+        format='GTiff'
+    )
 
-    del src_ds, dst_ds
+    flip_values_command = ' '.join([
```
forrestfwilliams, line 116: This could be a Python function like:
```python
def replace_geotiff_data(geotiff_path: str, stock_value: float) -> None:
    """Load a geotiff file, swap the non-zero data for all ones and resave."""
    ds = gdal.Open(geotiff_path, gdal.GA_Update)
    band = ds.GetRasterBand(1)
    data = band.ReadAsArray()
    data[data != 0 + 0j] = stock_value
    band.WriteArray(data)
    ds.FlushCache()
    ds = None
```

tests/test_water_mask.py
AndrewPlayer3: Since golden tests will check that the masks are correct, by proxy, my intent with the test was to ensure that the function would successfully acquire the data and create a product without failing. I understand that the fact network is required to run that test is probably not ideal, so I'll see if I can break this apart. It's also worth noting that the test with downloading only takes a second or two.

src/hyp3_isce2/water_mask.py
AndrewPlayer3: I ended up having to add a bit more to that function. I feel like the CLI version is cleaner and more consistent with the rest of the code. Is there a significant benefit to this method? I compared the two in terms of speed and they're about the same.

```python
def switch_mask_values(input_image: str, output_image: str, gdal_format: str):
    """Swap the 1's and 0's in the provided mask.
    
    Args:
        input_image: 
    """
    ds = gdal.Open(input_image, gdal.GA_Update)
    band = ds.GetRasterBand(1)
    data = band.ReadAsArray()
    tmp = np.copy(data)
    data[tmp == 0 + 0j] = 1
    data[tmp != 0 + 0j] = 0
    band.WriteArray(data)
    ds.FlushCache()
    ds = None
    system(f'gdal_translate {input_image} {output_image} -of {gdal_format}')
```

CHANGELOG.md
AndrewPlayer3: I'm don't think any functionality has been removed. The inputs and outputs are the same. The water masks were fed to isce2 as rasters previously. Could you please clarify what functionality you are referring to?

tests/test_water_mask.py
```diff
@@ -1,127 +1,63 @@
...
+        ['/vsicurl/https://asf-dem-west.s3.amazonaws.com/WATER_MASK/TILES/n15w100.tif']
+    )
+    case_2 = (
+        'tests/data/test_geotiff.tif',
+        ['/vsicurl/https://asf-dem-west.s3.amazonaws.com/WATER_MASK/TILES/n30w120.tif']
+    )
+    assert water_mask.get_tiles(case_1[0], tmp_path=str(tmp_path)) == case_1[1]
+    assert water_mask.get_tiles(case_2[0], tmp_path=str(tmp_path)) == case_2[1]
+
+
```
forrestfwilliams, line 55:
```suggestion

@pytest.mark('integration')
```

forrestfwilliams (approved):

PR #189, opened 2024-01-25
Add tests for burst.py and perform some mild refactoring

forrestfwilliams:


cirrusasf (commented):
Is the double quote a standard way? I remember we use single quote for string.

forrestfwilliams:
My mistake @cirrusasf, I had my auto formatter misconfigured. Thanks for the catch!

cirrusasf (approved):

PR #197, opened 2024-02-08
Breakdown readme

jacquelynsmale:


jacquelynsmale:
@forrestfwilliams I got this to work locally and looks reasonable with a quick lookover. I checked for basic blocks, but I'd love a review to check that 

- The order of the blocks is correct
- The breakdown of the blocks is correct (ex: that the InSAR processing is not its own block, that those subsections are ok being condensed into one block)
- Double-check for copypasta 

> 

>

src/hyp3_isce2/insar_tops_burst.py
```diff
@@ -194,7 +194,7 @@ def make_readme(
         'dem_pixel_spacing': '30 m',
         'apply_water_mask': apply_water_mask,
     }
-    content = hyp3_isce2.metadata.util.render_template('insar_burst/readme.md.txt.j2', payload)
+    content = hyp3_isce2.metadata.util.render_template('insar_burst/insar_burst_readme.md.txt.j2', payload)
```
jacquelynsmale, line 197:
```suggestion
    content = hyp3_isce2.metadata.util.render_template('insar_burst/insar_burst_base.md.txt.j2', payload)
```

PR #200, opened 2024-02-15
add description of water mask option in the README.md and update teh …

cirrusasf:
…CHANGELOG.md


forrestfwilliams:
Honestly I don't think we even need this PR. I think adding the extra info in the readme is kinda clunky and everything else is already fixed!

cirrusasf:
I agree with your idea. We don't need this PR.

PR #202, opened 2024-02-20
Remove GSHHG reference

forrestfwilliams:


src/hyp3_isce2/metadata/templates/insar_burst/insar_burst_base.md.txt.j2
```diff
@@ -235,7 +235,7 @@ had the water mask applied.
 
 The water mask is generated using data from OpenStreetMaps and/or ESA WorldCover depending on location. Areas within
 Canada, Alaska, and Russia are primarily covered by ESA WorldCover data, while the rest of the world is covered
-by OpenStreetMaps data. Water masks were previously generated from the GSHHG dataset.
+by OpenStreetMaps data.
```
hjkristenson, line 238: Is the description on lines 227-229 still correct? It seems like the new mask is much more aggressive with the small lakes (we didn't bother masking small lakes with the GSHHG; it was just the biggest inland waterbodies), and I'm not sure what the approach for landfast Antarctic ice is with the new datasets. @AndrewPlayer3 might have more insight on that.

hjkristenson (approved):
I left one comment (which I also just noted in the gamma PR, too) to double-check a different section of the water mask description, but if that's all still okay, the rest looks good.

src/hyp3_isce2/metadata/templates/insar_burst/insar_burst_base.md.txt.j2
forrestfwilliams: I guess it hangs on what you define as "major". I'm inclined to keep the language the same since Andrew did try to include comparable waterbodies. Luckily we'll be publishing the code for this new water mask, so the truly curious will be able to determine exactly what we did.

jacquelynsmale (approved):

Issue #212, opened 2024-06-16
hyp3_isce2_burst_merge tutorial failed at merging step


LiChiehLin:
Hi all,

I followed the tutorial notebook from this: 
https://github.com/ASFHyP3/hyp3-docs/blob/develop/docs/tutorials/hyp3_isce2_burst_merge.ipynb
and an error showed when it reached running hyp3_isce2 plugin 
```python
!python -m hyp3_isce2 ++process merge_tops_bursts $data_dir
```
Below is the error it showed.
![image](https://github.com/ASFHyP3/hyp3-docs/assets/86345753/38185144-5980-48ca-af09-b0abde058658)
I modified line 131 in `merge_tops_bursts.py` and it worked all the way to the end without any further errors.
```
start_utc = [
        datetime.datetime.strptime(result.properties['startTime'], '%Y-%m-%dT%H:%M:%SZ') for result in results
    ]
```
Is it possible the naming fashion had changed?
Thanks!

PR #216, opened 2024-06-28
Update time parsing in `merge-tops-bursts.py`

jacquelynsmale:
Update time parsing as suggested by [issue 212](https://github.com/ASFHyP3/hyp3-isce2/issues/212).


forrestfwilliams:
I also added `--overwrite` to the `gdal_calc.py` call in `water_mask.py`. Without this, test_water_mask.py would fail if you had previously run the tests locally because it wasn't able to overwrite the generated test files.

forrestfwilliams (approved):

PR #220, opened 2024-07-01
Fix shadowing of logging built-in

forrestfwilliams:


jtherrmann (approved):

.flake8
jtherrmann, line 1: Can we just delete this file?

jhkennedy: Yep!

forrestfwilliams: Ok, I'll delete it!

PR #231, opened 2024-08-13
Multi bursts

mfangaritav:
This is a draft for the new multi bursts workflow


src/hyp3_isce2/utils.py
```diff
@@ -1,133 +1,17 @@
...
-        for key, value in self.options.items():
-            gdal.SetConfigOption(key, value)
-
-    def __exit__(self, exc_type, exc_val, exc_tb):
-        for key, value in self._previous_options.items():
-            gdal.SetConfigOption(key, value)
-
-
-@dataclass
-class ParameterFile:
```
AndrewPlayer3, line 42: This gets called in `merge_tops_burst` along with `make_browse_image`. It might be worth leaving these functions in utils.py if we can.

AndrewPlayer3 (approved):

PR #239, opened 2024-08-16
Multi-Burst Fixes

AndrewPlayer3:
Changes burst workflow back to using `granules` for reference and secondary inputs. This supports the multi-burst workflow and the single-burst workflow.


mfangaritav (commented):
We should change back the README to drop the --reference and --secondary flags

mfangaritav (approved):

PR #245, opened 2024-08-26
Fix insar_tops workflow

forrestfwilliams:
Small reworking to remove a now unnecessary option that introduced a bug.

### DIFF
--- a/src/hyp3_isce2/insar_tops.py
+++ b/src/hyp3_isce2/insar_tops.py
@@ -49,12 +49,9 @@ def insar_tops(
     aux_cal_dir = Path('aux_cal')
     dem_dir = Path('dem')
 
-    if download:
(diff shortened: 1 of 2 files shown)


src/hyp3_isce2/insar_tops.py
```diff
@@ -49,12 +48,9 @@ def insar_tops(
     aux_cal_dir = Path('aux_cal')
     dem_dir = Path('dem')
 
-    if download:
```
jtherrmann, line 52: Should the `download: bool = True` param be removed from the `insar_tops` function?

forrestfwilliams: Ack, you're right. Removed.

jtherrmann (approved):

PR #256, opened 2025-01-06
Add `mypy` Workflow for Type Checking

AndrewPlayer3:
TODO:

- [x] Update the mypy action version pin after we release actions

### DIFF
--- a/.github/workflows/static-analysis.yml
+++ b/.github/workflows/static-analysis.yml
@@ -13,5 +13,4 @@ jobs:
 
   call-mypy-workflow:
     # Docs: https://github.com/ASFHyP3/actions
-    # TODO: update version tag after actions release
-    uses: ASFHyP3/actions/.github/workflows/reusable-mypy.yml@develop
+    uses: ASFHyP3/actions/.github/workflows/reusable-mypy.yml@v0.14.0
--- a/.github/workflows/static-analysis.yml
+++ b/.github/workflows/static-analysis.yml
@@ -10,3 +10,7 @@ jobs:
   call-ruff-workflow:
     # Docs: https://github.com/ASFHyP3/actions
     uses: ASFHyP3/actions/.github/workflows/reusable-ruff.yml@v0.12.0
+
+  call-mypy-workflow:
+    # Docs: https://github.com/ASFHyP3/actions
+    uses: ASFHyP3/actions/.github/workflows/reusable-mypy.yml@mypy
--- a/.github/workflows/static-analysis.yml
+++ b/.github/workflows/static-analysis.yml
@@ -13,4 +13,4 @@ jobs:
 
   call-mypy-workflow:
     # Docs: https://github.com/ASFHyP3/actions
-    uses: ASFHyP3/actions/.github/workflows/reusable-mypy.yml@mypy
+    uses: ASFHyP3/actions/.github/workflows/reusable-mypy.yml@0.14.0
--- a/.github/workflows/static-analysis.yml
+++ b/.github/workflows/static-analysis.yml
@@ -5,11 +5,11 @@ on: push
 jobs:
   call-secrets-analysis-workflow:
     # Docs: https://github.com/ASFHyP3/actions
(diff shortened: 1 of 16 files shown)


pyproject.toml
```diff
@@ -103,3 +103,9 @@ convention = "google"
 [tool.ruff.lint.isort]
 case-sensitive = true
 lines-after-imports = 2
+
```
jtherrmann, line 106: Add python version for mypy

pyproject.toml
```diff
@@ -103,3 +103,11 @@ convention = "google"
...
+[tool.mypy]
+python_version = "3.10"
```
jtherrmann, line 108: What happens if we use the same type specifier as the `requires-python` field above?
```suggestion
python_version = ">=3.10"
```

src/hyp3_isce2/burst.py
```diff
@@ -237,16 +245,28 @@ def spoof_safe(burst: BurstMetadata, burst_tiff_path: Path, base_path: Path = Pa
...
 
-    etree.ElementTree(burst.annotation).write(annotation_path / burst.annotation_name, **et_args)
-    etree.ElementTree(burst.calibration).write(calibration_path / burst.calibration_name, **et_args)
-    etree.ElementTree(burst.noise).write(calibration_path / burst.noise_name, **et_args)
-    etree.ElementTree(burst.manifest).write(safe_path / 'manifest.safe', **et_args)
+    for filepath, element in zip(filepaths, elements):
+        if element:
+            etree.ElementTree(element).write(filepath, encoding=encoding, xml_declaration=xml_dec)
+        else:
+            raise ValueError(f'Error writing burst metadata to {filepath}.')
```
jtherrmann, line 269: I think I prefer the simplicity of the `assert isinstance` approach that you used elsewhere, e.g. `assert isinstance(burst.annotation, etree._Element)`. Any reason not to use that method here? So we could stick closer to the original code (rather than using a `for` loop), but add an `assert isinstance` for each of the four elements.

jtherrmann (changes requested):

src/hyp3_isce2/burst.py
AndrewPlayer3: It was the first one I did before I realized adding proper exceptions would take too much time. I honestly think proper exceptions are better than asserts. I can change that part but getting rid of the loop is definitely worse.

```python
assert isinstance(burst.annotation, etree._Element)
assert isinstance(burst.calibration, etree._Element)
assert isinstance(burst.noise, etree._Element)
assert isinstance(burst.manifest, etree._Element)
etree.ElementTree(burst.annotation).write(annotation_path / burst.annotation_name, encoding=encoding, xml_declaration=xml_dec)
etree.ElementTree(burst.calibration).write(calibration_path / burst.calibration_name, encoding=encoding, xml_declaration=xml_dec)
etree.ElementTree(burst.noise).write(calibration_path / burst.noise_name,encoding=encoding, xml_declaration=xml_dec)
etree.ElementTree(burst.manifest).write(safe_path / 'manifest.safe', encoding=encoding, xml_declaration=xml_dec)
```

jtherrmann: Fair, I just want to be cautious about changing too much as part of adding mypy. Is this chunk of code covered by our tests?

.github/workflows/static-analysis.yml
```diff
@@ -5,8 +5,13 @@ on: push
...
 
   call-ruff-workflow:
     # Docs: https://github.com/ASFHyP3/actions
-    uses: ASFHyP3/actions/.github/workflows/reusable-ruff.yml@v0.12.0
+    uses: ASFHyP3/actions/.github/workflows/reusable-ruff.yml@v0.13.2
+
+  call-mypy-workflow:
+    # Docs: https://github.com/ASFHyP3/actions
+    # TODO: update version tag after actions release
+    uses: ASFHyP3/actions/.github/workflows/reusable-mypy.yml@develop
```
jtherrmann, line 17:
```suggestion
    uses: ASFHyP3/actions/.github/workflows/reusable-mypy.yml@v0.14.0
```

jtherrmann (approved):

PR #258, opened 2025-01-06
jth mypy changes

jtherrmann:
Changes:
- Downgrade some of our reusable actions from `v0.14.0` (not yet released)
- Remove some unnecessary `str()` casts (I confirmed with [`reveal_type`](https://mypy.readthedocs.io/en/latest/common_issues.html#reveal-type) that mypy already correctly infers those types as `str`)
- Change some usages of [`AttributeError`](https://docs.python.org/3.11/library/exceptions.html#AttributeError) to [`ValueError`](https://docs.python.org/3.11/library/exceptions.html#ValueError), because `AttributeError` specifically refers to Python object attributes, which isn't how we're using exceptions in this case.

### DIFF
--- a/.github/workflows/static-analysis.yml
+++ b/.github/workflows/static-analysis.yml
@@ -5,12 +5,13 @@ on: push
 jobs:
   call-secrets-analysis-workflow:
     # Docs: https://github.com/ASFHyP3/actions
-    uses: ASFHyP3/actions/.github/workflows/reusable-secrets-analysis.yml@v0.14.0
+    uses: ASFHyP3/actions/.github/workflows/reusable-secrets-analysis.yml@v0.13.2
 
   call-ruff-workflow:
     # Docs: https://github.com/ASFHyP3/actions
-    uses: ASFHyP3/actions/.github/workflows/reusable-ruff.yml@v0.14.0
+    uses: ASFHyP3/actions/.github/workflows/reusable-ruff.yml@v0.13.2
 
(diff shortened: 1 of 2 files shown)


.github/workflows/static-analysis.yml
```diff
@@ -5,12 +5,13 @@ on: push
...
   call-ruff-workflow:
     # Docs: https://github.com/ASFHyP3/actions
-    uses: ASFHyP3/actions/.github/workflows/reusable-ruff.yml@v0.14.0
+    uses: ASFHyP3/actions/.github/workflows/reusable-ruff.yml@v0.13.2
 
   call-mypy-workflow:
     # Docs: https://github.com/ASFHyP3/actions
-    uses: ASFHyP3/actions/.github/workflows/reusable-mypy.yml@0.14.0
+    # TODO: update version tag after actions release
+    uses: ASFHyP3/actions/.github/workflows/reusable-mypy.yml@develop
```
AndrewPlayer3, line 17: These will all be 14.0 after the release right? We have to do the actions release first anyway.

jtherrmann: Yeah, but I wanted those workflows (especially the mypy workflow) to run on our `hyp3-isce2` PRs so I could make sure I wasn't breaking anything. But, I'd be fine with reverting this change after we do the actions release.

AndrewPlayer3 (approved):

PR #274, opened 2025-02-28
Burst validation updates

jtherrmann:
Burst validation updates to go with https://github.com/ASFHyP3/hyp3/pull/2631

Changes include:
- Standardized around "scene" instead of "burst" to refer to scene names (I'm open to being persuaded otherwise)
- Improved some error messages to include more specific information about the problem
- Fixed some pycharm warnings (redundant parens, `multilook_position` variable shadows function name)
- Removed unused `swaths` param from `insar_tops_multi_burst`

TODO:

- [x] `TODO` items in code
- [x] Add/update tests
- [x] Midnight boundary edge case
- [x] Changelog

### DIFF
--- a/src/hyp3_isce2/burst.py
+++ b/src/hyp3_isce2/burst.py
@@ -444,10 +444,14 @@ def validate_bursts(reference: str | list[str], secondary: str | list[str]) -> N
     sec_datetimes = sorted(_burst_datetime(g) for g in secondary)
 
     if ref_datetimes[-1] - ref_datetimes[0] > timedelta(minutes=2):
-        raise ValueError('Reference scenes must fall within a 2-minute window in order to ensure they were collected during the same pass')
+        raise ValueError(
+            'Reference scenes must fall within a 2-minute window in order to ensure they were collected during the same pass'
+        )
 
     if sec_datetimes[-1] - sec_datetimes[0] > timedelta(minutes=2):
-        raise ValueError('Secondary scenes must fall within a 2-minute window in order to ensure they were collected during the same pass')
+        raise ValueError(
+            'Secondary scenes must fall within a 2-minute window in order to ensure they were collected during the same pass'
+        )
 
     if ref_datetimes[-1] >= sec_datetimes[0]:
         raise ValueError('Reference scenes must be older than secondary scenes')
--- a/src/hyp3_isce2/burst.py
+++ b/src/hyp3_isce2/burst.py
@@ -421,10 +421,6 @@ def validate_bursts(reference: str | list[str], secondary: str | list[str]) -> N
             f'Must provide the same number of reference and secondary scenes, got {len(reference)} reference and {len(secondary)} secondary'
         )
 
-    ref_num_swath_pol = [_num_swath_pol(g) for g in reference]
-    sec_num_swath_pol = [_num_swath_pol(g) for g in secondary]
-
-   
     for ref, sec in zip(reference, secondary):
         if _num_swath_pol(ref) != _num_swath_pol(sec):
             raise ValueError(
--- a/src/hyp3_isce2/burst.py
+++ b/src/hyp3_isce2/burst.py
@@ -392,6 +392,14 @@ def get_burst_params(scene_name: str) -> BurstParams:
     )
 
 
+def _num_swath_pol(scene: str) -> str:
+    parts = scene.split('_')
+    num = parts[1]
+    swath = parts[2]
+    pol = parts[4]
+    return '_'.join([num, swath, pol])
+
+
 def validate_bursts(reference: str | list[str], secondary: str | list[str]) -> None:
     """Check whether the reference and secondary bursts are valid.
 
@@ -441,14 +449,6 @@ def validate_bursts(reference: str | list[str], secondary: str | list[str]) -> N
 
     if ref_dates[0] >= sec_dates[0]:
         raise ValueError('Reference scenes must be older than secondary scenes')
-
-
-def _num_swath_pol(scene: str) -> str:
-    parts = scene.split('_')
-    num = parts[1]
-    swath = parts[2]
-    pol = parts[4]
-    return '_'.join([num, swath, pol])
 
 
 def load_burst_position(swath_xml_path: str, burst_number: int) -> BurstPosition:
--- a/src/hyp3_isce2/burst.py
+++ b/src/hyp3_isce2/burst.py
@@ -422,22 +422,22 @@ def validate_bursts(reference: str | list[str], secondary: str | list[str]) -> N
                 f'Number + swath + polarization identifier does not match for reference scene {reference[i]} and secondary scene {secondary[i]}'
             )
 
-    pols = sorted(set(g.split('_')[4] for g in reference))
+    pols = list(set(g.split('_')[4] for g in reference))
 
     if len(pols) > 1:
-        raise ValueError(f'Scenes must have the same polarization. Polarizations present: {", ".join(pols)}')
+        raise ValueError(f'Scenes must have the same polarization. Polarizations present: {", ".join(sorted(pols))}')
 
     if pols[0] not in ['VV', 'HH']:
         raise ValueError(f'{pols[0]} polarization is not currently supported, only VV and HH')
 
-    ref_dates = sorted(set(g.split('_')[3][:8] for g in reference))
-    sec_dates = sorted(set(g.split('_')[3][:8] for g in secondary))
+    ref_dates = list(set(g.split('_')[3][:8] for g in reference))
+    sec_dates = list(set(g.split('_')[3][:8] for g in secondary))
 
     if len(ref_dates) > 1:
-        raise ValueError(f'Reference scenes must be from a single date. Dates present: {", ".join(ref_dates)}')
+        raise ValueError(f'Reference scenes must be from a single date. Dates present: {", ".join(sorted(ref_dates))}')
 
     if len(sec_dates) > 1:
-        raise ValueError(f'Secondary scenes must be from a single date. Dates present: {", ".join(sec_dates)}')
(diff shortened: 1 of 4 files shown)


src/hyp3_isce2/burst.py
```diff
@@ -401,38 +401,54 @@ def validate_bursts(reference: str | list[str], secondary: str | list[str]) -> N
...
+    sec_num_swath_pol = [_num_swath_pol(g) for g in secondary]
 
-    ref_num_swath_pol = sorted(g.split('_')[1] + '_' + g.split('_')[2] + '_' + g.split('_')[4] for g in reference)
-    sec_num_swath_pol = sorted(g.split('_')[1] + '_' + g.split('_')[2] + '_' + g.split('_')[4] for g in secondary)
-    if ref_num_swath_pol != sec_num_swath_pol:
-        msg = 'The reference and secondary burst ID sets do not match.\n'
-        msg += f'    Reference IDs: {ref_num_swath_pol}\n'
-        msg += f'    Secondary IDs: {sec_num_swath_pol}'
-        raise ValueError(msg)
+    for i in range(len(reference)):
```
forrestfwilliams, line 419: I think it would be cleaner to define the `num_swath_pol` variables here within the for loop

src/hyp3_isce2/burst.py
```diff
@@ -401,38 +401,54 @@ def validate_bursts(reference: str | list[str], secondary: str | list[str]) -> N
...
+        if ref_num_swath_pol[i] != sec_num_swath_pol[i]:
+            raise ValueError(
+                f'Number + swath + polarization identifier does not match for reference scene {reference[i]} and secondary scene {secondary[i]}'
+            )
 
-    pols = list(set(g.split('_')[4] for g in reference + secondary))
+    pols = list(set(g.split('_')[4] for g in reference))
```
forrestfwilliams, line 425: You need to create `pols` using both reference and secondary. If you only use `reference`, you're not checking if the reference/secondary scenes share one single polarization.

src/hyp3_isce2/burst.py
```diff
@@ -401,38 +401,54 @@ def validate_bursts(reference: str | list[str], secondary: str | list[str]) -> N
...
 
     if len(pols) > 1:
-        raise ValueError(f'All bursts must have a single polarization. Polarizations present: {" ".join(pols)}')
+        raise ValueError(f'Scenes must have the same polarization. Polarizations present: {", ".join(sorted(pols))}')
 
     if pols[0] not in ['VV', 'HH']:
-        raise ValueError(f'{pols[0]} polarization is not currently supported, only VV and HH.')
+        raise ValueError(f'{pols[0]} polarization is not currently supported, only VV and HH')
 
     ref_dates = list(set(g.split('_')[3][:8] for g in reference))
```
forrestfwilliams, line 433: You didn't introduce this, but we don't account for cross-midnight burst sets here. Since we're updating the check I think it would be worth accounting for this edge case.

src/hyp3_isce2/burst.py
```diff
@@ -401,38 +401,54 @@ def validate_bursts(reference: str | list[str], secondary: str | list[str]) -> N
...
+
+    if len(sec_dates) > 1:
+        raise ValueError(f'Secondary scenes must be from a single date. Dates present: {", ".join(sorted(sec_dates))}')
 
     if ref_dates[0] >= sec_dates[0]:
-        raise ValueError('Reference granules must be older than secondary granules.')
+        raise ValueError('Reference scenes must be older than secondary scenes')
+
+
+def _num_swath_pol(scene: str) -> str:
```
forrestfwilliams, line 446: I'd prefer to see this function above `validate_bursts`. I generally expect functions to be above the functions that call them in a file.

src/hyp3_isce2/insar_tops_burst.py
```diff
@@ -219,7 +219,6 @@ def insar_tops_single_burst(
 def insar_tops_multi_burst(
     reference: list[str],
     secondary: list[str],
-    swaths: list = [1, 2, 3],
```
forrestfwilliams, line 222: Can you give me a bit more detail on this change? Why do we remove the `swaths` argument? Was it just not being used?

src/hyp3_isce2/burst.py
```diff
@@ -401,38 +401,54 @@ def validate_bursts(reference: str | list[str], secondary: str | list[str]) -> N
...
-    if ref_num_swath_pol != sec_num_swath_pol:
-        msg = 'The reference and secondary burst ID sets do not match.\n'
-        msg += f'    Reference IDs: {ref_num_swath_pol}\n'
-        msg += f'    Secondary IDs: {sec_num_swath_pol}'
-        raise ValueError(msg)
+    for i in range(len(reference)):
+        if ref_num_swath_pol[i] != sec_num_swath_pol[i]:
+            raise ValueError(
+                f'Number + swath + polarization identifier does not match for reference scene {reference[i]} and secondary scene {secondary[i]}'
+            )
```
williamh890, line 423:
```suggestion
   
    for ref, sec in zip(reference, secondary):
        if _num_swath_pol(ref) != _num_swath_pol(sec):
            raise ValueError(
                f'Number + swath + polarization identifier does not match for reference scene {ref} and secondary scene {sec}'
            )
```

src/hyp3_isce2/burst.py
jtherrmann: We've already checked that number+swath+pol is the same for each ref/sec pair, so we should only need to get the set of reference pols here, right?

src/hyp3_isce2/burst.py
jtherrmann: Done.

src/hyp3_isce2/insar_tops_burst.py
jtherrmann: Yep, it was unused. It's set explicitly in this function:

```python
swaths = list(set(int(granule.split('_')[2][2]) for granule in reference))
```

but the parameter value is never used. Also, `insar_tops_multi_burst` is only called in once place in this repo (at the bottom of this file), and `swaths` is not passed there.

Also note that using a mutable default argument such as `[1, 2, 3]` is discouraged, so PyCharm complains about that as well: https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

src/hyp3_isce2/burst.py
jtherrmann: Nice!

src/hyp3_isce2/burst.py
jtherrmann: Done.

src/hyp3_isce2/burst.py
jtherrmann: Ah, sorry, I missed this requirement in your earlier Mattermost message. Some clarifying questions:

- Is the requirement that the oldest reference scene is no more than 24 hours apart from the newest reference scene, and same for the secondary scenes?
- Should we require than the user provide the reference and secondary scenes sorted by datetime (oldest first, newest last)? Or does the order not matter as long as they're within the same 24-hour window?
- How does the 1 day apart rule change the "reference scenes must be older than secondary scenes" rule? Does there still need to be zero overlap between the reference time window and the secondary time window? (i.e. the newest reference scene must be older than the oldest secondary scene)
- Do the reference and secondary lists need to have similar datetime distributions, e.g. does it matter if all of the reference scenes are within a few minutes of each other, while the secondary scenes are spread evenly across the 24-hour window?

jtherrmann: I realized after I wrote all that the 24-hour requirement might be specific to your OPERA-DISP example, and not relevant to bursts. So I guess my first question should have been: what length of time window do the bursts need to fall within?

jtherrmann: Summarizing a live conversation, the requirements are:

- All of the reference scenes must be within a 2-min window (newest - oldest <= 2 min), and same for secondary scenes
- The time windows must be non-overlapping

src/hyp3_isce2/burst.py
forrestfwilliams: Yep, you're right

src/hyp3_isce2/burst.py
```diff
@@ -401,38 +414,43 @@ def validate_bursts(reference: str | list[str], secondary: str | list[str]) -> N
...
-        raise ValueError(f'{pols[0]} polarization is not currently supported, only VV and HH.')
+        raise ValueError(f'{pols[0]} polarization is not currently supported, only VV and HH')
+
+    ref_datetimes = sorted(_burst_datetime(g) for g in reference)
+    sec_datetimes = sorted(_burst_datetime(g) for g in secondary)
 
-    ref_dates = list(set(g.split('_')[3][:8] for g in reference))
-    sec_dates = list(set(g.split('_')[3][:8] for g in secondary))
+    if ref_datetimes[-1] - ref_datetimes[0] > timedelta(minutes=2):
+        raise ValueError('Reference scenes must fall within a 2-minute window')
```
forrestfwilliams, line 447:
```suggestion
        raise ValueError('Reference scenes must be collected during the same pass')
```

src/hyp3_isce2/burst.py
```diff
@@ -401,38 +414,43 @@ def validate_bursts(reference: str | list[str], secondary: str | list[str]) -> N
...
 
-    if len(ref_dates) > 1 or len(sec_dates) > 1:
-        raise ValueError('Reference granules must be from one date and secondary granules must be another.')
+    if sec_datetimes[-1] - sec_datetimes[0] > timedelta(minutes=2):
+        raise ValueError('Secondary scenes must fall within a 2-minute window')
```
forrestfwilliams, line 450:
```suggestion
        raise ValueError('Secondary scenes must be collected during the same pass')
```

forrestfwilliams (approved):

PR #280, opened 2025-03-07
Improve the Memory Efficiency of Water Masking

AndrewPlayer3:
This PR modifies `water_mask.py` such that `create_water_mask` is able to utilize the COG features of the water mask files in AWS. The improvement comes from subsetting the water mask tiles to the projection window of the input image. This means that GDAL only has to work with a small portion of the data, rather than the multiple GBs that the full water masks may take when uncompressed.


tests/data/water_mask_output_info.txt
jhkennedy, line 1: @AndrewPlayer3 can you explain why the test output image and info has changed?

jhkennedy (commented):
This looks good to me, overall. I'm just curious about the test info changes.

src/hyp3_isce2/water_mask.py
```diff
@@ -76,58 +85,45 @@ def create_water_mask(
...
 
-    The water mask is assembled from OpenStreetMap data.
+    The water masks are assembled from OpenStreetMap and ESA WorldCover data.
 
     Shoreline data is unbuffered and pixel values of 1 indicate land touches the pixel and 0 indicates there is no
     land in the pixel.
 
     Args:
-        input_image: Path for the input GDAL-compatible image
+        input_image: Path for the input GDAL-compatible image (needs to be in EPSG:4326).
```
jhkennedy, line 94: Should we raise or warn if this is not the case?

tests/data/water_mask_output_info.txt
AndrewPlayer3, line 1: For the record, this change in test info is due a few things. The input GeoTIFF was converted to EPSG:4326 from EPSG:32615 (new function requirement and this is what the input DEM is). There is also a change in the size, and slight shift, of the image due to the water mask function better matching the pixel size and spatial extent of the input image. Note that the input image is 10x10, not 11x10.

AndrewPlayer3: This change in test info is due a few things. The input GeoTIFF was converted to EPSG:4326 from EPSG:32615 (new function requirement and this is what the input DEM is). There is also a change in the size, and slight shift, of the image due to the water mask function better matching the pixel size and spatial extent of the input image. Note that the input image is 10x10, not 11x10.

src/hyp3_isce2/water_mask.py
AndrewPlayer3: I could, although this isn't a "user facing" function and should always be fed the EPSG:4326 DEM by the process. Is it possible that DEM may change to a different SRS in the future?

jhkennedy: We're using DEM Stitcher on the back end, which always returns EPSG:4326, so unless the DEM stitcher interface chances, that should be a valid assumption. That is a third-part package, so I can't decide if we should be more cautious or not here. I'm fine leaving it as is.

jhkennedy (approved):

Issue #165, opened 2023-11-21
`ValueError: There should only be 2 VRT files` error for `insar_isce_burst` workflow during preprocessing


asjohnston-asf:
### The bug
The `insar_isce_burst` workflow (and INSAR_ISCE_TEST jobs run via HyP3) potentially fail with this error during ISCE preprocessing:
`ValueError: There should only be 2 VRT files in the reference and secondary directories, this indicates there is likely a bug in the region of interest generation.`

### To Reproduce
I've only observed this error for this pair of bursts (so far):
```
S1_165626_IW2_20230708T005451_VV_00E5-BURST
S1_165626_IW2_20230720T005452_VV_CD3B-BURST
```

The workflow fails when run at any of `20x4`, `10x2`, or `5x1` looks.

Example failed jobs:
- https://hyp3-api.asf.alaska.edu/jobs/a4f827aa-dc5f-4fa8-882a-8369a44ee8f7
- https://hyp3-api.asf.alaska.edu/jobs/005e68d5-04d3-4eda-aa63-dbaf96ab855e
- https://hyp3-api.asf.alaska.edu/jobs/793aa9cb-def5-460a-b891-0626edc55cce

AndrewPlayer3:
I've observed this for more bursts. The issue is that the lower left corner of the intersection of the burst footprints may lie in multiple bursts in one of the SLCs. 

To illustrate, here is a visualization of the footprints for the following bursts:
S1_133436_IW2_20211214T001412_VV_153A-BURST
S1_133436_IW2_20220218T001451_VV_C384-BURST

<img width="957" alt="Image" src="https://github.com/user-attachments/assets/c10f3628-7322-4e65-b457-dc3bfb0e5c15" />

Note that "Sec Burst 4" is the burst immediately south of "S1_133436_IW2_20220218T001451_VV_C384-BURST", and it overlaps the lower left corner of the intersection of the two desired bursts. ISCE2 crops the burst list to ones that contain the ROI, thus it returns 2 secondary bursts for the ROI given by the lower left corner of the intersection.

AndrewPlayer3:
I'm still seeing a failure for jobs using burst id `290225`. I haven't seen any other failing scenes. Note that it is not a regression from #286, as the failure occurs with the previous ROI calculation as well. So that fix should still reduce the number of incidents of this bug.

[...]

PR #283, opened 2025-03-12
Fix for Erroneous `ValueError: There should only be 2 VRT files`

AndrewPlayer3:
This PR contains a fix for #165. 

According to the definition of ROI in `get_region_of_interest`, the region of interest is the lower left for descending scenes, and the upper right for ascending scenes. However, the current implementation uses the lower left for ascending scenes, and the lower right for descending scenes. This PR sets the corners correctly according to the provided definition for ascending scenes; however, it will implement the descending scenes as the **upper left** corner instead of the lower left one, as the lower left corner does not uniquely define bursts in descending orbits.

I've tested this locally, and have verified that previously broken burst pairs now work. I have also tested that previously working burst pairs still work. If we end up with the same issue in the future, we may need to implement trying multiple corners until one uniquely subsets the bursts within their respective SLCs.

For the record, the bounds are in the following format: [min_lon, min_lat, max_lon, max_lat].


jhkennedy (approved):
This looks good to me, but it'd be worth @forrestfwilliams weighing in as well.

AndrewPlayer3:
> This looks good to me, but it'd be worth @forrestfwilliams weighing in as well.

I think he is out till Friday. Do you think its worth waiting to do anything with this PR, or do you think its alright to merge and get the golden tests running? We can wait for his approval to release. I'm not sure if nows a bad time to have something sitting in develop.

jhkennedy:
Ah yes, let's go ahead and merge to develop, and get his input before we release.

PR #294, opened 2025-04-04
Use `burst2safe` for Single Burst Jobs to Fix Multiple VRT Bug

AndrewPlayer3:
This PR addresses #165. Setting the ISCE2 ROI to the corner of a burst will not always lead to a single-burst being subsetted by ISCE2. This happens regardless of the choice of corner. Using `burst2safe` avoids this problem by creating a safe file with only the desired burst in it.

Locally tested to verify that the processing is working.

TODO:
- [x] Remove functions that are no longer used
- [x] Address mypy errors
- [x] Fix the unit tests
- [x] Make sure that the `merge_tops_burst` workflow still functions correctly

### DIFF
--- a/src/hyp3_isce2/packaging.py
+++ b/src/hyp3_isce2/packaging.py
@@ -384,7 +384,8 @@ def make_parameter_file(
         multilook_position: Burst position for multilooked radar geometry products
         dem_name: Name of the DEM that is use
         dem_resolution: Resolution of the DEM
-
+        reference_safe: The name of the reference SAFE file (only necessary for single-burst)
+        secondary_safe: The name of the reference SAFE file (only necessary for single-burst)
     returns:
         None
     """
--- a/src/hyp3_isce2/packaging.py
+++ b/src/hyp3_isce2/packaging.py
@@ -384,8 +384,8 @@ def make_parameter_file(
         multilook_position: Burst position for multilooked radar geometry products
         dem_name: Name of the DEM that is use
         dem_resolution: Resolution of the DEM
-        reference_safe: The name of the reference SAFE file (only necessary for single-burst)
-        secondary_safe: The name of the reference SAFE file (only necessary for single-burst)
+        reference_safe: The name of the reference SAFE directory. Optional; will look for a `.SAFE` directory matching `reference_scene`  if not provided.
+        secondary_safe: The name of the reference SAFE directory. Optional; will look for a `.SAFE` directory matching `secondary_scene`  if not provided.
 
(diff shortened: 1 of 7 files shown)


src/hyp3_isce2/packaging.py
```diff
@@ -397,8 +401,10 @@ def make_parameter_file(
     else:
         ref_tag = reference_scene[-4::]
         sec_tag = secondary_scene[-4::]
-    reference_safe = [file for file in os.listdir('.') if file.endswith(f'{ref_tag}.SAFE')][0]
-    secondary_safe = [file for file in os.listdir('.') if file.endswith(f'{sec_tag}.SAFE')][0]
+        reference_safe = [file for file in os.listdir('.') if file.endswith(f'{ref_tag}.SAFE')][0]
+        secondary_safe = [file for file in os.listdir('.') if file.endswith(f'{sec_tag}.SAFE')][0]
```
jhkennedy, line 405:
```suggestion
    if reference_safe is None:
        reference_safe = [file for file in os.listdir('.') if file.endswith(f'{ref_tag}.SAFE')][0]
        secondary_safe = [file for file in os.listdir('.') if file.endswith(f'{sec_tag}.SAFE')][0]
```
I think I'd just let these be optional and look for them if they are `None`. (could check both but checking reference should be sufficient)

src/hyp3_isce2/packaging.py
```diff
@@ -382,6 +384,8 @@ def make_parameter_file(
         multilook_position: Burst position for multilooked radar geometry products
         dem_name: Name of the DEM that is use
         dem_resolution: Resolution of the DEM
+        reference_safe: The name of the reference SAFE file (only necessary for single-burst)
+        secondary_safe: The name of the reference SAFE file (only necessary for single-burst)
```
jhkennedy, line 388:
```suggestion
        reference_safe: The name of the reference SAFE directory. Optional; will look for a `.SAFE` directory matching `reference_scene`  if not provided.
        secondary_safe: The name of the reference SAFE directory. Optional; will look for a `.SAFE` directory matching `secondary_scene`  if not provided.
```

jhkennedy (approved):
This looks good to merge and start golden testing to me!

I left one minor take it or leave it suggestion. 

And, I think there's quite a bit of condensing/eliminating duplicate code that could still be done. But, I think that's for a later PR to make review and testing easier.

jhkennedy (dismissed):

jhkennedy (approved):

PR #302, opened 2025-04-18
Fix bug in metadata fetching

forrestfwilliams:


src/hyp3_isce2/packaging.py
```diff
@@ -402,18 +401,23 @@ def make_parameter_file(
...
-        reference_safe = [file for file in os.listdir('.') if file.endswith(f'{ref_tag}.SAFE')][0]
-        secondary_safe = [file for file in os.listdir('.') if file.endswith(f'{sec_tag}.SAFE')][0]
-
-    assert reference_safe and secondary_safe
-
-    ref_annotation_path = f'{reference_safe}/annotation/'
-    ref_annotation = [file for file in os.listdir(ref_annotation_path) if os.path.isfile(ref_annotation_path + file)][0]
+        reference_safe_list = list(Path.cwd().glob(f'*{ref_tag}.SAFE'))
+        assert len(reference_safe_list) == 1, f'Expected one reference SAFE, found {len(reference_safe_list)}'
+        reference_safe = reference_safe[0]
```
asjohnston-asf, line 406: First item of the list, yes? Not first character of the safe?
```suggestion
        reference_safe = reference_safe_list[0]
```

src/hyp3_isce2/packaging.py
```diff
@@ -402,18 +401,23 @@ def make_parameter_file(
...
+        secondary_safe_list = list(Path.cwd().glob(f'*{sec_tag}.SAFE'))
+        assert len(secondary_safe_list) == 1, f'Expected one secondary SAFE, found {len(secondary_safe_list)}'
+        secondary_safe = secondary_safe_list[0]
+
+    ref_annotation_path = reference_safe / 'annotation'
+    swath = reference_scene.split('_')[2].lower()
```
asjohnston-asf, line 412: `reference_scene` may be an SLC SAFE name, not a Burst Granule name, so we can't depend on the filename to extract a swath number (and multi-burst jobs may span multiple swaths).

asjohnston-asf:
Confirmed mintpy relies on `Heading` and `Slantrangenear` from the hyp3 parameter file: https://github.com/insarlab/MintPy/blob/main/src/mintpy/prep_hyp3.py#L60

asjohnston-asf:
@forrestfwilliams `make_parameter_file` expects `topsProc.xml` and `topsApp.xml` in the current working directory to fetch perpendicular baseline and unwrapper type / filter strength respectively. Can we parameterize those paths or data values as well?

asjohnston-asf (approved):

Issue #288, opened 2025-03-22
Multi-burst jobs should accept reference and secondary lists in any order with respect to burst number + swath + pol


jtherrmann:
**Jira:** https://asfdaac.atlassian.net/browse/TOOL-3566

*Note: The above link is accessible only to members of ASF.*

----------------------------------------------------------------------------------------------------

In the `validate_bursts` function in [burst.py](https://github.com/ASFHyP3/hyp3-isce2/blob/develop/src/hyp3_isce2/burst.py), we require that the reference and secondary lists be given in the same order with respect to the number + swath + pol identifier. Per @asjohnston-asf, we should allow them to be passed in any order and simply check that the *set* of IDs is the same.

Also, as part of updating `validate_bursts`, it would be good to add a comment to remind ourselves that the reference and secondary list requirements must be kept in sync with the burst product guide (and we'll need to update the guide as part of making this change).

PR #318, opened 2025-05-21
Accept any order for multi-burst

mfangaritav:
This PR fixes issue #288


src/hyp3_isce2/burst.py
```diff
@@ -266,10 +266,12 @@ def validate_bursts(reference: list[str], secondary: list[str]) -> None:
...
 
-    for ref, sec in zip(reference, secondary):
-        if _num_swath_pol(ref) != _num_swath_pol(sec):
+    secs_nsp = [_num_swath_pol(sec) for sec in secondary]
+
+    for ref in reference:
+        if _num_swath_pol(ref) not in secs_nsp:
             raise ValueError(
-                f'Number + swath + polarization identifier does not match for reference scene {ref} and secondary scene {sec}'
+                f'Number + swath + polarization identifier does not match for reference scene {ref} and secondary scenes'
```
AndrewPlayer3, line 274: This could probably be worded a bit better for someone who doesn't know what we might be talking about. Maybe something like
```suggestion
                f'Orbit number, swath, or polarization for reference scene {ref} does not match any secondary scenes.'
```

tests/test_burst.py
```diff
@@ -103,7 +103,7 @@ def test_validate_bursts_num_swath_pol():
 
     with pytest.raises(
         ValueError,
-        match=r'^Number \+ swath \+ polarization identifier does not match for reference scene S1_000001_IW2_20200101T000001_VV_0000\-BURST and secondary scene S1_000002_IW2_20200201T000001_VV_0000\-BURST$',
+        match=r'^Number \+ swath \+ polarization identifier does not match for reference scene S1_000001_IW2_20200101T000001_VV_0000\-BURST and secondary scenes',
```
AndrewPlayer3, line 106: It's probably worth adding one more test case where the orders are different.

AndrewPlayer3 (changes requested):

tests/test_burst.py
mfangaritav: I changed the order of the bursts in some of current tests

AndrewPlayer3 (approved):

Issue #211, opened 2024-06-17
Scenes over water fail with `Failed to compute statistics, no valid pixels found in sampling` when `apply_water_mask=True`


asjohnston-asf:
INSAR_ISCE_BURST jobs fail with errors such as `RuntimeError: S1_320206_IW1_20200901_20200913_VV_INT80_77F1/S1_320206_IW1_20200901_20200913_VV_INT80_77F1_unw_phase.tif, band 1: Failed to compute statistics, no valid pixels found in sampling.` during the geocoding step when a user processes a pair of scenes entirely over water and applies water masking.

For example, this pair is off the Florida coast:
```
S1_320206_IW1_20200901T232050_VV_8BB7-BURST
S1_320206_IW1_20200913T232051_VV_DD3A-BURST
```

An INSAR_ISCE_BURST jobs succeeds when `apply_water_mask=False`: https://hyp3-api.asf.alaska.edu/jobs/17ce7f24-ba52-4c44-abde-6a339fab7a87
But fails when `apply_water_mask=True`: https://hyp3-api.asf.alaska.edu/jobs/f3c721ee-827e-434b-a71a-50d1fe6d6c8d

![Screenshot from 2024-06-17 14-59-23](https://github.com/ASFHyP3/hyp3-isce2/assets/17994518/acec3471-7c4a-4e24-8e18-ad2bbc33d692)

[f3c721ee-827e-434b-a71a-50d1fe6d6c8d.log](https://github.com/user-attachments/files/15878211/f3c721ee-827e-434b-a71a-50d1fe6d6c8d.log)

lidar-nevermore:
I think this is the same bug as https://github.com/ASFHyP3/hyp3-isce2/issues/247
The lonlat coordinates are wrong, so that pixels in the water mask cannot be correctly transformed  from wgs84 to radar coordinates.

PR #319, opened 2025-05-23
Allow burst in water with water mask

mfangaritav:
This PR fixes issue #211


jhkennedy (dismissed):
Looks good! I confirmed that applying the water mask to the scene in #211 does not work in develop, but does with this PR. The produced browse image is all transparent, which matches the underlying all-nodata geotiff. 

I left one take it or leave it suggestion. 

And it might be worth adding a test for this behavior  simmilar to this test:
https://github.com/ASFHyP3/hyp3-isce2/blob/develop/tests/test_utils.py#L56-L61

tests/test_utils.py
```diff
@@ -59,6 +59,11 @@ def test_make_browse_image():
     utils.make_browse_image(input_tif, output_png)
     assert open(output_png, 'rb').read() == open('tests/data/test_browse_image.png', 'rb').read()
     os.remove(output_png)
+    input_empty = 'tests/data/test_empty.tif'
```
jhkennedy, line 62:
```suggestion

    input_empty = 'tests/data/test_empty.tif'
```
Nice to put a little white space between the different test blocks, or, even put it into it's own test like:
```
def test_make_browse_image_geotiff_all_nodata():
```

mfangaritav: No problem, I made a new test for it

PR #334, opened 2025-06-23
resolve importlib.metadata deprecation warning

asjohnston-asf:
running `python -m hyp3_isce2 insar_tops_burst --help` currently generates this deprecation warning:

```
/home/asjohnston/src/hyp3-isce2/src/hyp3_isce2/__main__.py:51: DeprecationWarning: SelectableGroups dict interface is deprecated. Use select.
  eps = entry_points()['hyp3']
```

See https://docs.python.org/3/library/importlib.metadata.html#entry-points , in particular:

> Changed in version 3.12: The “selectable” entry points were introduced in importlib_metadata 3.6 and Python 3.10. Prior to those changes, entry_points accepted no parameters and always returned a dictionary of entry points, keyed by group. With importlib_metadata 5.0 and Python 3.12, entry_points always returns an EntryPoints object. See [backports.entry_points_selectable](https://pypi.org/project/backports.entry_points_selectable/) for compatibility options.

### DIFF
--- a/src/hyp3_isce2/__main__.py
+++ b/src/hyp3_isce2/__main__.py
@@ -46,10 +46,7 @@ def main():
             UserWarning,
         )
 
-    # NOTE: Cast to set because of: https://github.com/pypa/setuptools/issues/3649
-    # NOTE: Will need to update to `entry_points(group='hyp3', name=args.process)` when updating to python 3.10
-    eps = entry_points(group='hyp3')
-    (process_entry_point,) = {process for process in eps if process.name == args.process}
+    process_entry_point = list(entry_points(group='hyp3', name=args.process))[0]
 
     if args.omp_num_threads:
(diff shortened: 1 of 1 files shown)


src/hyp3_isce2/__main__.py
```diff
@@ -46,10 +46,7 @@ def main():
             UserWarning,
         )
 
-    # NOTE: Cast to set because of: https://github.com/pypa/setuptools/issues/3649
-    # NOTE: Will need to update to `entry_points(group='hyp3', name=args.process)` when updating to python 3.10
-    eps = entry_points()['hyp3']
-    (process_entry_point,) = {process for process in eps if process.name == args.process}
+    process_entry_point = list(entry_points(group='hyp3', name=args.process))[0]
```
jhkennedy, line 49: Is the return not hashable anymore? I'd prefer to use the set for deduplication instead of just selecting the first one, so that you get an error if there are multiple, different entry points returned.

PR #398, opened 2025-11-19
Release v4.1.2

mfangaritav:
This is the golden test for multiburst: https://github.com/ASFHyP3/hyp3-testing/actions/runs/19512650796

### DIFF
--- a/src/hyp3_isce2/insar_tops_multi_burst.py
+++ b/src/hyp3_isce2/insar_tops_multi_burst.py
@@ -15,7 +15,7 @@ from hyp3_isce2 import packaging
 from hyp3_isce2.burst import validate_bursts
 from hyp3_isce2.insar_tops import insar_tops
 from hyp3_isce2.logger import configure_root_logger
-from hyp3_isce2.utils import get_multiburst_prefix, make_browse_image
+from hyp3_isce2.utils import get_publish_name, get_publish_prefix, make_browse_image
 
 
 gdal.UseExceptions()
@@ -161,6 +161,12 @@ def main():
         help='Additionally, publish products to this bucket. Necessary credentials must be provided '
         'via the `PUBLISH_ACCESS_KEY_ID` and `PUBLISH_SECRET_ACCESS_KEY` environment variables.',
     )
+    parser.add_argument(
+        '--publish-prefix',
+        type=str,
+        default=None,
+        help='Prefix for the bucket where the products will be published',
+    )
 
     args = parser.parse_args()
 
@@ -188,5 +194,9 @@ def main():
         packaging.upload_product_to_s3(product_dir, output_zip, args.bucket, args.bucket_prefix)
 
     if args.publish_bucket:
-        prefix, s3_name = get_multiburst_prefix(output_zip)
+        if args.publish_prefix:
+            prefix = args.publish_prefix
+        else:
+            prefix = get_publish_prefix(output_zip)
+        s3_name = get_publish_name(output_zip) 
         packaging.upload_file_to_s3_with_publish_access_keys(output_zip, args.publish_bucket, prefix, s3_name)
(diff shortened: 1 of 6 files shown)


src/hyp3_isce2/utils.py
```diff
@@ -393,3 +393,15 @@ def get_projection(srs_wkt) -> str:
...
+def get_multiburst_prefix(fpath: Path) -> tuple:
+    # files will go to the burst set folder
+    burst_set = str(fpath.name).replace('-', '_')
+    keep = [burst_set.split('_')[i] for i in [1, 2, 3, 4, 9]]
+    subkeep = str(fpath.name).split('_')[0:-1] + ['0000']
+    dir_name = '_'.join(keep)
+    s3_name = '_'.join(subkeep) + '.zip'
+    prefix = f'multiburst_products/{dir_name}'
+
+    return prefix, s3_name
```
jhkennedy: This function could be written in a way that is easier read, and there's a little bit going on in this function I don't really understand. 

I'd probably also call it `get_publish_prefix_and_name` (long), `get_publish_info` or `get_publish_location` to clearly indicate this is just for publication.

A more readable version I think looks like:

```python
def get_publish_location(product: Path) -> tuple[str, str]:
    parts = product.name.split('-')

    _, path, bursts, _, _, _, _, product_type, _ = parts
    prefix = f'multiburst_products/{path}_{bursts.replace("-", "_")}_{product_type}'
    
    parts[-1] = '0000.zip'  # non-unique product identifier
    name = '_'.join(parts)
    
    return prefix, name
```

And may be worth splitting into two `get_publish_prefix` and `get_publish_name` functions.

What I don't understand:

2. Why change the dashes between bursts to underscores in the prefix?

3. Why change the random hash in the file name? I assume you're intentionally trying to clobber existing products?

src/hyp3_isce2/insar_tops_multi_burst.py
```diff
@@ -179,3 +186,7 @@ def main():
 
     if args.bucket:
         packaging.upload_product_to_s3(product_dir, output_zip, args.bucket, args.bucket_prefix)
+
+    if args.publish_bucket:
+        prefix, s3_name = get_multiburst_prefix(output_zip)
```
jhkennedy:
```suggestion
        publish_name = get_publish_name(output_zip)
        if args.publish_prefix is None:
            publish_prefix = get_publish_prefix(output_zip)
        
```
you could add a `--publish-prefix` argument that would allow you to customize the prefix for non-volcsarvatory use. We will possibly need to add this to ITS_LIVE at some point as well.

src/hyp3_isce2/insar_tops_multi_burst.py
```diff
@@ -154,6 +154,13 @@ def main():
     )
     parser.add_argument('--bucket', help='AWS S3 bucket HyP3 for upload the final product(s)')
     parser.add_argument('--bucket-prefix', default='', help='Add a bucket prefix to product(s)')
+    parser.add_argument(
+        '--publish-bucket',
+        type=str,
+        default=None,
+        help='Additionally, publish products to this bucket. Necessary credentials must be provided '
+        'via the `PUBLISH_ACCESS_KEY_ID` and `PUBLISH_SECRET_ACCESS_KEY` environment variables.',
+    )
```
jhkennedy, line 163: It may also be worth adding a `--publish-prefix` option that would override the computed prefix, if provided.

src/hyp3_isce2/utils.py
mfangaritav: No problem. I'll change the function. About your questions:

2. No particular reason, I just don't like dashes and underscores in the same name but I can leave it as the original (I could add the S1 at the beginning too).

3. Yes, my idea is to overwrite existing products and avoid duplicates.

mfangaritav:
This is the golden test after the recent changes: https://github.com/ASFHyP3/hyp3-testing/actions/runs/19546457876

jhkennedy (approved):

